PR #4, opened 2025-01-31
build: release pipeline

gregorydemay:
([XC-284](https://dfinity.atlassian.net/browse/XC-284)): Use [release-plz](https://release-plz.dev/) to start automatizing the release process:

1. Job can be manually triggered to create a new PR containing the new versions and changelog for the next release. An example of such PR is #83.
2. An opened PR should trigger the normal CI flow (in particular the end-to-end tests with the deployment of the canister wasm built reproducibly). However, a [limitation](https://release-plz.dev/docs/github/token) of the default `GITHUB_TOKEN` in actions is that it cannot trigger other workflows. An easy workaround is to close/re-open the PR to trigger the necessary jobs.

A future PR will use `release-plz release` to publish the required crates on crates.io and create a Github release containing the wasm and the Candid file.

[XC-284]: https://dfinity.atlassian.net/browse/XC-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


libs/client/Cargo.toml
```diff
@@ -22,8 +22,8 @@ solana-clock = { workspace = true }
 solana-pubkey = { workspace = true }
 solana-signature = { workspace = true }
 solana-transaction-status-client-types = { workspace = true }
-sol_rpc_types = { path = "../types" }
+sol_rpc_types = { version = "0.1.0", path = "../types" }
```
gregorydemay, line 25: @ninegua: We need to set a version here so that it can be published by `cargo publish`

release-plz.toml
```diff
@@ -0,0 +1,52 @@
...
+changelog_update = true # enable changelog update for `package_a`
+git_release_enable = true # enable GitHub/Gitea releases for `package_a`
+publish = false # disable `cargo publish` for `package_a`
+
+[[package]]
+name = "basic_solana"
+release = false # don't process this package
+
+[[package]]
+name = "canlog"
```
ninegua, line 43: Is this going to be replaced by the released version?

ninegua (approved):

release-plz.toml
gregorydemay: @ninegua Yes! This is in #84

PR #52, opened 2025-03-31
feat: cycles cost

gregorydemay:
Add endpoints to the SOL RPC canister to estimate the cycle cost of a given request. The proposed API is as a follows:

1. A caller can pass the exact same request parameters to the cycles cost estimation endpoint as to the endpoint that would actually execute the request so that a precise cycles cost estimation can be returned for that particular request.
2. Since the endpoints on the SOL RPC canister use different types for the parameters, there is a  dedicated `_CycleCost` endpoint for every Solana endpoint supported by the canister.
3. The idea is that the caller could use those endpoints to establish a baseline and then do the other calls for that endpoint with some additional cycles to account for minor changes in request parameters (unused cycles are reimbursed).

Implementation details:

1. Introduce the concept of  `MultiRpcRequest`, which is a JSON-RPC request aimed at multiple URLs (corresponding to Solana JSON-RPC providers). This replaces the previous clients used in the canister (`CandidRpcClient` and `SolRpcClient`).
2. This allows to implement once a `cycles_cost` method for all `MultuRpcRequest` that we will support.

### DIFF
--- a/canister/src/rpc_client/mod.rs
+++ b/canister/src/rpc_client/mod.rs
@@ -18,7 +18,7 @@ use canhttp::{
     CyclesChargingPolicy, CyclesCostEstimator, MaxResponseBytesRequestExtension,
     TransformContextRequestExtension,
 };
-use http::Request;
+use http::{Request, Response};
 use ic_cdk::api::management_canister::http_request::CanisterHttpRequestArgument as IcHttpRequest;
 use ic_cdk::api::management_canister::http_request::TransformContext;
 use serde::{de::DeserializeOwned, Serialize};
@@ -227,15 +227,12 @@ impl<Params, Output> MultiRpcRequest<Params, Output> {
 
         let client = service_request_builder()
             .service_fn(extract_request)
-            .map_err(|e: HttpClientError| RpcError::from(e))
-            .map_response(|r| r.into_body());
+            .map_err(RpcError::from)
+            .map_response(Response::into_body);
 
         let (requests, errors) = requests.into_inner();
-        if !errors.is_empty() {
-            return Err(errors
-                .into_values()
-                .next()
-                .expect("BUG: errors is not empty"));
+        if let Some(error) = errors.into_values().next() {
+            return Err(error);
         }
 
         let (_client, results) = canhttp::multi::parallel_call(client, requests).await;


canister/src/main.rs
```diff
@@ -82,45 +83,61 @@ async fn get_slot(
...
+    match MultiRpcRequest::get_slot(
         source,
-        config.map(RpcConfig::from),
-        rounding_error,
+        config.unwrap_or_default(),
+        params.unwrap_or_default(),
     ) {
-        Ok(client) => client.get_slot(params.unwrap_or_default()).await,
-        Err(err) => Err(err).into(),
+        Ok(request) => process_result(RpcMethod::GetSlot, request.send_and_reduce().await),
```
lpahlavi, line 91: nit: I think I would add the `RpcMethod` as a field of the `request` and make this match statement a part of the `send_and_reduce` method (potential rename?) since it's the same for all calls. Then this should look more similar to the `get_slot_cycles_cost` method which is easier to read IMO.

Potential output transformations (e.g. `serde_json::Value` -> `String` for raw requests) can still be performed on the `MultiRpcRequest` through `map`.

canister/src/rpc_client/mod.rs
```diff
@@ -78,103 +159,113 @@ impl SolRpcClient {
...
+        ) -> Result<http::Response<IcHttpRequest>, HttpClientError> {
+            Ok(http::Response::new(request))
+        }
+
+        let num_providers = self.providers.sources.len();
+        let requests = self.create_json_rpc_requests();
+
+        let client = service_request_builder()
+            .service_fn(extract_request)
+            .map_err(|e: HttpClientError| RpcError::from(e))
```
lpahlavi, line 202: nit: `.map_err(RpcError::from).map_response(Response::into_body)`

lpahlavi (commented):
Thanks a lot for the PR @gregorydemay! Looks very good, just some minor comments on my side.

canister/sol_rpc_canister.did
```diff
@@ -239,7 +243,9 @@ service : (InstallArgs,) -> {
 
   // Call the Solana `getSlot` RPC method and return the resulting slot.
   getSlot : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (MultiGetSlotResult);
+  getSlotCyclesCost : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (RequestCostResult) query;
 
-  // Make a generic RPC request that sends the given json_rpc_payload.
-  request : (RpcSources, opt RpcConfig, json_rpc_paylod: text) -> (MultiRequestResult)
+  // Make a generic JSON RPC request that sends the given json_rpc_payload.
+  jsonRequest : (RpcSources, opt RpcConfig, json_rpc_payload: text) -> (MultiRequestResult);
```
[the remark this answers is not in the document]
gregorydemay: > I believe in some spots we call this "raw", in others "generic" and here "json"

Good catch. It makes sense to aim for consistency. I picked `jsonRequest` but other alternatives are also welcomed (cc @ninegua )

canister/src/main.rs
gregorydemay: > I think I would add the RpcMethod as a field of the request

Good idea! I actually think we don't really need `RpcMethod` because `MultiRpcRequest` already contains  `JsonRpcRequest` which by definition must have the specified `method. This would change a bit the semantics of the metrics `inconsistent_responses` since we would not distinguish between Candid-supported methods and the ones coming from the raw `jsonRequest` endpoint but I'm not sure this distinction is work it.

Since this PR is already quite big, I suggest to do that in a follow-up one. Ok?

canister/src/main.rs
```diff
@@ -82,45 +83,61 @@ async fn get_slot(
     config: Option<GetSlotRpcConfig>,
     params: Option<GetSlotParams>,
 ) -> MultiRpcResult<Slot> {
-    let rounding_error = config
-        .as_ref()
-        .and_then(|c| c.rounding_error)
-        .map(RoundingError::from);
-    match CandidRpcClient::new_with_rounding_error(
+    match MultiRpcRequest::get_slot(
```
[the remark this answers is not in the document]
gregorydemay: I think this will change in a follow-up PR due to the comment below.

canister/src/rpc_client/mod.rs
```diff
@@ -78,103 +159,113 @@ impl SolRpcClient {
...
+                .next()
+                .expect("BUG: errors is not empty"));
+        }
+
+        let (_client, results) = canhttp::multi::parallel_call(client, requests).await;
+        let (requests, errors) = results.into_inner();
+        if !errors.is_empty() {
+            return Err(errors
+                .into_values()
+                .next()
```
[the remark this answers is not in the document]
gregorydemay: good point, much cleaner that way!

canister/src/rpc_client/mod.rs
gregorydemay: Thanks!

gregorydemay (commented):
Thanks @lpahlavi and @ninegua for the feedback! Should be ready for another round.

canister/sol_rpc_canister.did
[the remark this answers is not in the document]
lpahlavi: `jsonRequest` is fine for me, thank you for addressing this!

canister/src/main.rs
lpahlavi: Yes, I agree. I think that's a good way to do it. I actually initially used the `JsonRpcRequest` method name for the metrics and then backtracked but I think it is probably the better option. I also think it's fine to have the metrics for generic JSON requests separated by which RPC method is called, this way we might even see if many users are calling RPC methods we don't support and which specific ones.

A follow-up PR is fine by me. Thanks for taking care of this!

lpahlavi (approved):
Thanks for addressing the feedback and rebasing the PR! LGTM!

PR #55, opened 2025-04-02
build: migrate dependencies to `solana-sdk` repository

lpahlavi:
### ❓ Why this change?

Previously, we relied on patched dependencies from the [`anza-xyz/agave`](https://github.com/anza-xyz/agave) repository. However, the Solana SDK has since been migrated to [`anza-xyz/solana-sdk`](https://github.com/anza-xyz/solana-sdk), making the old patch target obsolete.

Additionally, our workspace `Cargo.toml` specified patched dependencies using semver-compatible version ranges, such as:

```toml
solana-account-decoder-client-types = "2.2.0"
```

This works fine during `cargo build`, because patched versions from the fork are prioritized. But when running `cargo publish`, Cargo ignores the patches and resolves the latest semver-compatible upstream version instead (e.g., `v2.3.2`), which may introduce breaking changes. For example, in `v2.3.1` of the `solana-account-decoder-client-types` crate, a new (optional) `cost_units` field was added to `UiTransactionStatusMeta`, causing compilation failures at publish time.

### ✅ What’s the fix?
* 🔄 Patch all Solana SDK crates from the new `anza-xyz/solana-sdk` fork
* 📌 Pin all patched Solana versions using `=` to avoid semver drift:
     ```toml
     solana-account-decoder-client-types = "=2.3.3"
    ```
This ensures reproducible builds and a smooth `cargo publish` process for the SOL RPC canister crate.


Cargo.toml
```diff
@@ -103,26 +105,23 @@ inherits = "release"
...
-solana-reward-info = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-rpc-client-api = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-rpc-client-nonce-utils = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-signer = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-signature = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-transaction = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-transaction-context = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-transaction-error = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
-solana-transaction-status-client-types = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
+solana-account = { git = "https://github.com/dfinity/solana-sdk", rev = "46ca4e22a312a418cbbf14296305f750ec122d9d" }
```
gregorydemay, line 108: nit: I thinking making a tag (e.g., `46ca4e2-js-feature-flag`) would help for traceability reasons

Cargo.toml
```diff
@@ -59,27 +59,29 @@ serde = { version = "1.0.219", features = ["derive"] }
...
-solana-pubkey = "2.2.0"
-solana-reward-info = "2.2.0"
-solana-rpc-client-api = "2.2.0"
-solana-rpc-client-nonce-utils = "2.2.0"
-solana-signature = "2.2.0"
-solana-signer = "2.2.0"
-solana-transaction = "2.2.0"
-solana-transaction-error = "2.2.0"
-solana-transaction-status-client-types = "2.2.0"
+solana-account = "=2.2.1"
```
gregorydemay, line 62: understanding question: is it on purpose that they have different exact versions? E.g., `solana-account` uses `2.2.1` but `solana-clock` `2.2.2`, even thought they are in the same repo?

libs/client/CHANGELOG.md
```diff
@@ -5,9 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.0.0] - 2025-07-04
```
gregorydemay, line 8: I would not do the Changelog yet, since this should be taken care of (and probably adapted) from the release job, which is going to open a PR for this. It probably needs to be adapted since initially I thought that using 1 version for the whole repo was a good idea (which it was not 😄 )

gregorydemay (commented):
Thanks @lpahlavi for this PR! Some minor comments.

Cargo.toml
lpahlavi: Actually yes! The different crate versions have diverged quite a bit since they migrated to the the new `anza-xyz/solana-sdk` repository. Essentially, the versions here correspond to the current versions in the forked repository, which is based on `sdk@v2.3.1`.

gregorydemay (approved):

PR #57, opened 2025-04-07
build: use Dfinity fork

gregorydemay:
As a temporary solution until anza-xyz/solana-sdk#117 is merged upstream, use the fix from lpahlavi/agave@fa0bacd5508c1e443db79a67d470c537fa558de0 on the fork [dfinity/agave](https://github.com/dfinity/agave).

### DIFF
--- a/canister/scripts/provision.sh
+++ b/canister/scripts/provision.sh
@@ -8,6 +8,8 @@ set -u # or set -o nounset
 : "$ALCHEMY_MAINNET_API_KEY"
 : "$ALCHEMY_DEVNET_API_KEY"
 : "$ANKR_MAINNET_API_KEY"
+: "$DRPC_DEVNET_API_KEY"
+: "$DRPC_MAINNET_API_KEY"
 : "$ANKR_DEVNET_API_KEY"
 : "$HELIUS_MAINNET_API_KEY"
 : "$HELIUS_DEVNET_API_KEY"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -80,16 +80,16 @@ zeroize = { version = "1.8", features = ["zeroize_derive"] }
...
-solana-hash = { git = "https://github.com/lpahlavi/agave.git" }
-solana-instruction = { git = "https://github.com/lpahlavi/agave.git" }
-solana-keypair = { git = "https://github.com/lpahlavi/agave.git" }
-solana-message = { git = "https://github.com/lpahlavi/agave.git" }
-solana-nonce = { git = "https://github.com/lpahlavi/agave.git" }
-solana-program = { git = "https://github.com/lpahlavi/agave.git" }
-solana-pubkey = { git = "https://github.com/lpahlavi/agave.git" }
-solana-signature = { git = "https://github.com/lpahlavi/agave.git" }
-solana-transaction = { git = "https://github.com/lpahlavi/agave.git" }
-solana-system-interface = { git = "https://github.com/lpahlavi/solana-program-system.git" }
```
mducroux, line 95: Is it ok that this is not included?

mducroux (approved):

Cargo.toml
gregorydemay: If it builds without it and the tests are green, then it's ok 😄

lpahlavi: Yes, good catch, it seems like this isn't used anywhere anymore. That's great, one less repo to fork!

lpahlavi (approved):

PR #59, opened 2025-04-08
feat: add `sendTransaction` RPC method

lpahlavi:
(XC-294) Add support for the Solana`sendTransaction` RPC method.


Cargo.toml
```diff
@@ -95,5 +99,7 @@ solana-nonce = { git = "https://github.com/dfinity/agave", tag = "323039e-js-fea
 solana-program = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
 solana-pubkey = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
 solana-rpc-client-api = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
+solana-signer = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
 solana-signature = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
 solana-transaction = { git = "https://github.com/dfinity/agave", tag = "323039e-js-feature-flag" }
+solana-system-interface = { git = "https://github.com/lpahlavi/solana-program-system.git" }
```
lpahlavi, line 105: @gregorydemay This was removed in your PR, but actually it is needed to create transactions (specifically to create instructions to add to the transactions). I realized this is needed to build the example canister but we don't have any tests for this (yet) so we didn't realize. Since I need to create transactions in the integration tests for `sendTransactions`, I added this back. We can think if we want to create a fork for this repository owned by DFINITY as well while my PR in the corresponding repository is merged.

Another option would be to just re-implement the required functionality ourselves since we just need `solana_system_interface::instruction::transfer` (so far) but it might also be useful for end-users to be able to create transactions within a canister.

WDYT?

canister/src/rpc_client/sol_rpc/tests.rs
```diff
@@ -111,6 +111,16 @@ mod normalization_tests {
         );
     }
 
+    #[test]
+    fn should_normalize_send_transaction_response() {
```
gregorydemay, line 115: nit: basically it should leave the transaction ID untouched so you could maybe turn it into a proptest?

integration_tests/tests/solana_test_validator.rs
```diff
@@ -94,8 +98,52 @@ async fn should_not_get_account_info() {
...
+        PathBuf::from(var("CARGO_MANIFEST_DIR").unwrap()).join("keypair1.json"),
+    )
+    .unwrap();
+    let recipient = Keypair::read_from_file(
+        PathBuf::from(var("CARGO_MANIFEST_DIR").unwrap()).join("keypair2.json"),
+    )
+    .unwrap();
+
+    let blockhash = setup.solana_client.get_latest_blockhash().unwrap();
+    let instruction = system_instruction::transfer(&sender.pubkey(), &recipient.pubkey(), 1_000);
```
gregorydemay, line 115: understanding question: shouldn't get the sender some sol before doing the transaction? I'm guessing this doesn't pass preflight.

integration_tests/tests/tests.rs
```diff
@@ -374,6 +376,48 @@ mod get_slot_tests {
...
+                .mock_sequential_json_rpc_responses::<3>(
+                    200,
+                    json!({
+                        "id": first_id,
+                        "jsonrpc": "2.0",
+                        "result": "2vC221MDR312jrFzh5TRnMfUCHrCiG4cBuzHmagdgrQSsdLHaq65uJVLCWmubw4FkBDUxhRpQma785MpMwRS6ob7",
+                    }),
+                )
+                .build()
+                .send_transaction(SendTransactionParams {
```
gregorydemay, line 399: nit: It would be useful to be able to instantiate `SendTransactionParams` directly from a string without having to set all the other fields to `None`, similar to what was done for [GetAccountInfoParams](https://github.com/dfinity/sol-rpc-canister/blob/d8e5fba78afe89952f0878760652c9cadfb89aad/libs/types/src/solana/mod.rs#L111)

integration_tests/tests/tests.rs
```diff
@@ -374,6 +376,48 @@ mod get_slot_tests {
...
+                    max_retries: None,
+                    min_context_slot: None,
+                })
+                .send()
+                .await
+                .expect_consistent();
+
+            assert_eq!(
+                results,
+                Ok(Signature::from_str("2vC221MDR312jrFzh5TRnMfUCHrCiG4cBuzHmagdgrQSsdLHaq65uJVLCWmubw4FkBDUxhRpQma785MpMwRS6ob7").unwrap())
```
gregorydemay, line 413: nit: use a variable for `"2vC221MDR312jrFzh5TRnMfUCHrCiG4cBuzHmagdgrQSsdLHaq65uJVLCWmubw4FkBDUxhRpQma785MpMwRS6ob7"` to make it clear it's the same one as for the mocks.

libs/types/src/solana/mod.rs
```diff
@@ -1,11 +1,76 @@
...
+
+impl SendTransactionParams {
+    /// Returns `true` if all of the optional config parameters are `None` and `false` otherwise.
+    pub fn is_default_config(&self) -> bool {
+        self.encoding.is_none()
+            && self.skip_preflight.is_none()
+            && self.preflight_commitment.is_none()
+            && self.max_retries.is_none()
+            && self.min_context_slot.is_none()
+    }
```
gregorydemay, line 42: nit: I would use the destructuring pattern to catch the addition of any optional field
```suggestion
    pub fn is_default_config(&self) -> bool {
        let SendTransactionParams {
            transaction: _,
            encoding,
            skip_preflight,
            preflight_commitment,
            max_retries,
            min_context_slot,
        } = &self;
        encoding.is_none()
            && skip_preflight.is_none()
            && preflight_commitment.is_none()
            && max_retries.is_none()
            && min_context_slot.is_none()
    }
```

gregorydemay (commented):
Some minor comments but otherwise LGTM!

libs/types/src/solana/mod.rs
```diff
@@ -1,11 +1,76 @@
...
+}
+
+impl From<solana_transaction::Transaction> for SendTransactionParams {
+    fn from(transaction: solana_transaction::Transaction) -> Self {
+        fn encode(transaction: solana_transaction::Transaction) -> String {
+            bs58::encode(bincode::serialize(&transaction).expect("Failed to serialize transaction"))
```
ninegua, line 48: understanding question: why is this using `bincode`? Is it required by solana specification (if there is such a thing)?

Cargo.toml
lpahlavi: Thanks!

canister/sol_rpc_canister.did
```diff
@@ -238,6 +238,33 @@ type AccountEncoding = variant {
...
+// The parameters for a Solana `sendTransaction` RPC method call.
+type SendTransactionParams = record {
+    // Fully-signed transaction, as encoded string.
+    transaction: text;
+    // Encoding format for the transaction.
+    encoding: opt SendTransactionEncoding;
+    // When true, skip the preflight transaction checks. Default: false.
+    skipPreflight: opt bool;
+    // Commitment level to use for preflight. See Configuring State Commitment. Default finalized.
+    preflightCommitment: opt text;
```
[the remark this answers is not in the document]
lpahlavi: Yes, good point! Done.

canister/src/rpc_client/sol_rpc/tests.rs
lpahlavi: Good idea! Done.

integration_tests/tests/solana_test_validator.rs
lpahlavi: Good question, you are right but I already added this t o the CI pipeline so we can re-use these accounts in various integration tests. The accounts for these two keypairs are both airdroped 10 SOL [here](https://github.com/dfinity/sol-rpc-canister/blob/553d5a6bba0a0cc20878741e8f41f4d7ffd94dc9/.github/workflows/ci.yml#L102).

integration_tests/tests/tests.rs
lpahlavi: Good point. It's already possible to do that from a `solana_transaction::Transaction` (similar to how `GetAccountInfoParams` can be instantiated from a `solana_pubkey::Pubkey`) which is not that straightforward to build from a string. I realized `solana_transaction::Transaction` implements `Default` though so I used it here to simplify things.

We could also consider implementing `From<String>` for both `GetAccountInfoParams` and `SendTransactionParams` but I'm not sure it would be so useful outside of tests.

libs/types/src/solana/mod.rs
lpahlavi: Good idea! Done.

libs/types/src/solana/mod.rs
lpahlavi: Good question. The thing is that the RPC method actually takes a _serialized_ transaction (in base58 format) as parameter. However, for some reason, it seems to get the serialized transaction from a `solana_transaction::Transaction`, one has to manually serialize it using `bincode`. This is basically how the Rust Solana SDK is implemented. I'm not sure there is another way to get the serialized transaction otherwise. Does this make sense?

lpahlavi (commented):
Thanks a lot for the review @gregorydemay and @ninegua! I've addressed your feedback and this should now be ready for another round of reviews.

integration_tests/tests/tests.rs
gregorydemay: >  realized solana_transaction::Transaction implements Default though so I used it here to simplify things.

nice!

libs/types/src/solana/mod.rs
gregorydemay: > I'm not sure there is another way to get the serialized transaction otherwise.

I think there are 2 issues at hand:

1. Serialize the transaction.
2. Encode the bytes of the serialized transaction

For the 1st, `bincode` seems indeed necessary and is used [here](https://github.com/anza-xyz/agave/blob/9d99617ff0b7fbcbf335bfd7425654d2a5b14c1d/rpc-client/src/nonblocking/rpc_client.rs#L4646).

For the second I think we should not use base58, since this is deprecated as noted [here](https://solana.com/de/docs/rpc/http/sendtransaction) but rather use base64 as [here](https://github.com/anza-xyz/agave/blob/9d99617ff0b7fbcbf335bfd7425654d2a5b14c1d/rpc-client/src/nonblocking/rpc_client.rs#L4650).
WDYT?

ninegua: Thanks for the explanation! This seems rather strange though, because Rust SDK isn't the only RPC client, and I can't believe the binary format of a solana transaction depends on bincode, which is not a standard format AFAIK.

libs/types/src/solana/mod.rs
```diff
@@ -1,11 +1,76 @@
...
+/// A Solana base58-encoded [transaction ID](https://solana.com/docs/references/terminology#transaction-id).
+pub type TransactionId = String;
+
+/// The parameters for a Solana [`sendTransaction`](https://solana.com/docs/rpc/http/sendtransaction) RPC method call.
+#[derive(Debug, Clone, Deserialize, Serialize, CandidType)]
+pub struct SendTransactionParams {
+    /// Fully-signed transaction, as encoded string.
+    pub transaction: String,
+    /// Encoding format for the transaction.
+    pub encoding: Option<SendTransactionEncoding>,
```
gregorydemay, line 17: We probably don't want those 2 fields to be freely modifiable because they are co-dependent: setting `transaction` to a base64-encoded String and the appropriate `encoding` should no longer mutate (since changing the `encoding` to base58 without changing the `transaction` would be an error). Would it be possible to make those 2 fields private an create a constructor for the 2 required parameters?

libs/types/src/solana/mod.rs
ninegua: ChatGPT says Solana transaction uses a serialization format called Borsh, and Rust has borsh and borsh-derive. Not sure how true is that though, haven't verified.

gregorydemay: > Solana transaction uses a serialization format called Borsh, and Rust has borsh and borsh-derive.

I have seen both :-) I think newer Solana types use `Borsh` while old ones use `bincode`. See for example the docs on [`Instruction`](https://docs.rs/solana-instruction/2.2.1/src/solana_instruction/lib.rs.html#135-137)

ninegua: Yes, it does seem that they are actually just using bincode for RPC. Even the Javascript SDK tries to [decode/encode things](https://github.com/anza-xyz/kit/blob/10286701ddf0f6dcc04623b59f9513e36cf8654e/packages/transactions/src/codecs/transaction-codec.ts#L54) the way bincode does, I'm amazed...

lpahlavi: I spent way more time than I'd like to admit looking through the Solana docs and asking various LLMs where this is documented, but I think this is a good instance of "the code is the documentation" unfortunately... I also ended up looking at the Javascript SDK and came to the same conclusion as you @ninegua.

@gregorydemay I think it's fair enough to use base 64 for the serialization. The only caveat is that this means by default we need to include a config object in the request which makes it a little bit bigger, though it's only a matter of a few bytes. WDYT?

gregorydemay: @lpahlavi since base58 is marked as deprecated I would use base64, even if that means we must have a config object (depending on the transaction the resulting request could still be smaller since base64 encoding is more space efficient)

integration_tests/tests/solana_test_validator.rs
lpahlavi: > 1. I think it would help for readability (everything in one place) if this setup was part of the Rust test itself. AFAIK this should be possible with the Solana client and the [request_airdrop](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.request_airdrop) method.

I think that's a fair point. I changed the setup a bit so that everything is done in this file. Note that this means we need to wait for the Airdrop to actually get confirmed before we send the transaction. Until now the Solana client from the SDK was using the default value for the commitment level in all its requests which is `finalized` (makes sense for prod but not for tests) meaning 15s until the Airdrop is "valid". I changed the commitment level to `processed` so that everything is "confirmed" faster.

> 2. Additionally, I think we should test there that the sol balance of the users changed as expected to be sure that both transactions went through.

I actually overhauled the test a bit since it doesn't make sense to send the transaction using the Solana client. Transactions should only be sent once. Instead I added some asserts as you suggested to make sure the funds were transferred, as well as an assert that the returned transaction ID corresponds to a confirmed signature.

lpahlavi (commented):
@gregorydemay @ninegua Thank you for your feedback! I believe this should now be ready for another round of reviews.

integration_tests/tests/solana_test_validator.rs
```diff
@@ -102,42 +115,40 @@ async fn should_not_get_account_info() {
...
+        .send()
+        .await
+        .expect_consistent()
+        .unwrap();
+
+    // Make sure the funds were sent from the sender to the recipient
+    let sender_balance_after = setup.get_account_balance(&sender.pubkey());
+    let recipient_balance_after = setup.get_account_balance(&recipient.pubkey());
+
+    assert!(recipient_balance_after >= recipient_balance_before + transaction_amount);
```
gregorydemay, line 147: nit: why not having strict equality here?

integration_tests/tests/solana_test_validator.rs
```diff
@@ -102,42 +115,40 @@ async fn should_not_get_account_info() {
...
+
+    let transaction_amount = 1_000;
+    let instruction =
+        system_instruction::transfer(&sender.pubkey(), &recipient.pubkey(), transaction_amount);
     let transaction = Transaction::new_signed_with_payer(
         &[instruction],
         Some(&sender.pubkey()),
         &[&sender],
-        blockhash,
+        setup.get_latest_blockhash(),
```
gregorydemay, line 128: nit: maybe add a TODO for XC-289: to get the block hash via getSlot+getBlock

libs/types/src/solana/mod.rs
```diff
@@ -32,30 +33,57 @@ pub struct SendTransactionParams {
...
+            && skip_preflight.is_none()
+            && preflight_commitment.is_none()
+            && max_retries.is_none()
+            && min_context_slot.is_none()
+    }
+
+    /// The transaction being sent as an encoded string.
+    pub fn get_transaction(&self) -> String {
+        self.transaction.clone()
+    }
```
gregorydemay, line 72: nit: using something by reference and cloning is an anti-pattern. Let the caller decide whether cloning is necessary
```suggestion
    pub fn get_transaction(&self) -> &str {
        &self.transaction
    }
```

libs/types/src/solana/mod.rs
```diff
@@ -32,30 +33,57 @@ pub struct SendTransactionParams {
...
-            transaction: encode(transaction),
-            encoding: None,
-            skip_preflight: None,
-            preflight_commitment: None,
-            max_retries: None,
-            min_context_slot: None,
-        }
+        Self::from_encoded_transaction(
+            BASE64_STANDARD
+                .encode(bincode::serialize(&transaction).expect("Failed to serialize transaction")),
```
gregorydemay, line 84: nit: Implementation of the `From` [trait](https://doc.rust-lang.org/std/convert/trait.From.html) should never panic (ok for test code, but that's not the case here)

> **Note: This trait must not fail**. The From trait is intended for perfect conversions. If the conversion can fail or is not perfect, use [TryFrom](https://doc.rust-lang.org/std/convert/trait.TryFrom.html).

I think it makes sense to implement `TryFrom` and let it bubble up to an `RpcError::ValidationError` to the caller.

libs/types/src/solana/mod.rs
gregorydemay: > I left the encoding field optional, even though we never actually leave it None since this also affects the Candid interface.

I think it's ok here to leave as Optional

gregorydemay (approved):
Thanks a lot @lpahlavi for the changes! Left some minor comments but otherwise LGTM! Also ok to do a follow-up PR if you prefer

PR #61, opened 2025-04-09
refactor: use method from JSON-RPC request for metric

gregorydemay:
Follow-up on #52 to use the `method` specified by the JSON-RPC request for metrics purposes instead of having a separate enum.


lpahlavi (approved):
Thanks @gregorydemay!

ninegua (approved):

PR #63, opened 2025-04-10
docs: Add basic README instructions

gregorydemay:
Add README instructions to help projects getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,9 @@ The SOL RPC canister reaches the Solana JSON-RPC providers using [HTTPS outcalls
    to canonicalize the response seen by each node before doing consensus may alleviate the problem.
    For example, `getSlot` rounds by default the received slot by 20, therefore artificially increasing the slot time seen by each node to 8s to allow them reaching consensus with some significantly higher probability.
    The reason why such a canonicalization strategy does not work for [`getLatestBlockhash`](https://solana.com/de/docs/rpc/http/getlatestblockhash) is that the result is basically a random-looking string of fixed length.
+4. There are therefore two options to send a transaction on Solana using the SOL RPC canister
+   1. Use a [durable nonce](https://solana.com/de/developers/guides/advanced/introduction-to-durable-nonces) instead of a blockhash.
+   2. Retrieve a recent blockhash by first retrieving a recent slot with `getSlot` and then getting the block (which includes the blockhash) with `getBlock`.
 
 ## Reproducible Build
 
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![Internet Computer portal](https://img.shields.io/badge/InternetComputer-grey?logo=internet%20computer&style=for-the-badge)](https://internetcomputer.org)
+[![DFinity Forum](https://img.shields.io/badge/help-post%20on%20forum.dfinity.org-blue?style=for-the-badge)](https://forum.dfinity.org/t/sol-rpc-canister/41896)
+[![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg?logo=apache&style=for-the-badge)](LICENSE)
+
 # SOL RPC canister
 
 > [!IMPORTANT]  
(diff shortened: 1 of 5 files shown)


libs/client/README.md
```diff
@@ -1,2 +1,7 @@
+[![Internet Computer portal](https://img.shields.io/badge/InternetComputer-grey?logo=internet%20computer&style=for-the-badge)](https://internetcomputer.org)
+[![DFinity Forum](https://img.shields.io/badge/help-post%20on%20forum.dfinity.org-blue?style=for-the-badge)](https://forum.dfinity.org/t/sol-rpc-canister/41896)
+[![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg?logo=apache&style=for-the-badge)](LICENSE)
+
 # Crate `sol_rpc_client`
 
+Rust library to interact with the SOL RPC canister from a canister. See the crate documentation for more details.
```
lpahlavi, line 7: nit: This is actually the crate documentation for the `sol_rpc_client` so I guess this could be a bit confusing, no? Perhaps see the top-level documentation or something like that.

libs/types/README.md
```diff
@@ -1,2 +1,7 @@
+[![Internet Computer portal](https://img.shields.io/badge/InternetComputer-grey?logo=internet%20computer&style=for-the-badge)](https://internetcomputer.org)
+[![DFinity Forum](https://img.shields.io/badge/help-post%20on%20forum.dfinity.org-blue?style=for-the-badge)](https://forum.dfinity.org/t/sol-rpc-canister/41896)
+[![GitHub license](https://img.shields.io/badge/license-Apache%202.0-blue.svg?logo=apache&style=for-the-badge)](LICENSE)
+
 # Crate `sol_rpc_types`
 
+This crate defines types for interacting with the SOL RPC canister. See the crate documentation for more details.
```
lpahlavi, line 7: nit: same comment as for `libs/client/README.md`

README.md
```diff
@@ -7,3 +11,115 @@
...
+### From the command line
+
+#### Prerequisites:
+
+* [Install](https://internetcomputer.org/docs/building-apps/developer-tools/dev-tools-overview#dfx) `dfx`.
+* [Cycles wallet](https://internetcomputer.org/docs/building-apps/canister-management/cycles-wallet) with some cycles to pay for requests.
+
+#### Example with [`getSlot`](https://solana.com/de/docs/rpc/http/getslot)
+
+To get the last `finalized` slot on Solana Mainnet
```
lpahlavi, line 51: nit: I think I would either leave out the RPC config (i.e. `null`) or explain what it does. If you leave it out, it would also be more closely aligned with the Rust example below.

README.md
```diff
@@ -7,3 +11,115 @@
...
+
+Full examples are available in the [examples](examples) folder and additional code snippets are also available in the [`sol_rpc_client`](libs/client/README.md) crate.
+
+## Limitations
+
+The SOL RPC canister reaches the Solana JSON-RPC providers using [HTTPS outcalls](https://internetcomputer.org/https-outcalls) and are therefore subject to the following limitations:
+1. The contacted providers must support IPv6.
+2. Some Solana RPC endpoint cannot be supported. This is the case for example for [`getLatestBlockhash`](https://solana.com/de/docs/rpc/http/getlatestblockhash).
+   The reason is that an HTTPs outcalls involves an HTTP request from each node in the subnet and has therefore a latency in the order of a few seconds. 
+   This can be problematic for endpoints with fast changing responses, such as [`getLatestBlockhash`](https://solana.com/de/docs/rpc/http/getlatestblockhash), since in this case nodes will not be able to reach a consensus.
```
lpahlavi, line 103: nit: You could maybe mention here that `getSlot` is a bit special (don't need to get into the details of how) and can be used as an alternative e.g. combined with `getBlock`. I could just see that some readers would be confused how we support `getSlot` when it suffers from a very similar problem to `getLatestBlockhash`.

lpahlavi (approved):
Thanks for this PR @gregorydemay! LGTM, just some very minor comments.

libs/client/README.md
lpahlavi: Yes I was confused by "crate documentation". Maybe then it would make sense to say "Rust documentation" or something similar? Maybe it's just me 🙈

README.md
gregorydemay: Good catch, I don't want to have a `null` parameter since it's not really clear what it means when your read this the first time, so I added the choice of `Equality` also in the Rust example which was missing.

README.md
gregorydemay: good idea, I added more details. PTAL.

libs/client/README.md
gregorydemay: Ha I see. I added some links to the future place on https://docs.rs/ where the Rust docs should be. Hopefully this should be less confusing now.

gregorydemay (commented):
Thanks @lpahlavi for the feedback!

README.md
```diff
@@ -7,3 +11,129 @@
...
+* [Contributing](#contributing)
+* [License](#license)
+
+## Features
+
+* **No single point of failure**:  Each request will by default query 3 distinct Solana JSON-RPC providers and aggregate their results.
+* **Configurable consensus strategy**: Choose how responses from multiple providers are aggregated depending on the needs of your application, e.g., 3-out-of-5 meaning that 5 providers will be queried and the overall response will be successful if at least 3 do agree (equality).
+* **Pay directly in cycles**: No need to take care of API keys, each request can be paid for by attaching cycles.
+* **Bring your own**: 
+    * A Solana RPC method is not supported? There is an endpoint to send any JSON-RPC request.
```
mducroux, line 33: It could be helpful to specify which one.

README.md
```diff
@@ -7,3 +11,129 @@
...
+    * Missing your favorite Solana JSON-RPC provider? You can specify your own providers.
```
mducroux, line 34: Couldn't we include the way to specify the provider here?

README.md
```diff
@@ -7,3 +11,129 @@
...
+
+## Usage
+
+The SOL RPC canister runs on the [fiduciary subnet](https://internetcomputer.org/docs/current/concepts/subnet-types#fiduciary-subnets) with the following principal: [`tghme-zyaaa-aaaar-qarca-cai`](https://dashboard.internetcomputer.org/canister/tghme-zyaaa-aaaar-qarca-cai).
```
mducroux, line 38: This link returns a 502 error for me. Maybe you could include https://internetcomputer.org/docs/building-apps/developing-canisters/deploy-specific-subnet#fiduciary-subnets or https://dashboard.internetcomputer.org/subnet/pzp6e-ekpqk-3c5x7-2h6so-njoeq-mt45d-h3h6c-q3mxf-vpeq5-fk5o7-yae

mducroux (approved):
LGTM, only minor comments from my side.

README.md
gregorydemay: Good idea. Done.

README.md
gregorydemay: Good idea. I wouldn't go into too much details here, but I think having a hint so that people know what to look for is sufficient.

README.md
gregorydemay: Good catch. It's due to some renaming from the dev docs that broke many links (their redirect doesn't work)

gregorydemay (commented):
Thanks @mducroux for the feedback!

ninegua (approved):

README.md
```diff
@@ -7,3 +11,129 @@
...
+
+* [Install](https://internetcomputer.org/docs/building-apps/developer-tools/dev-tools-overview#dfx) `dfx`.
+* [Cycles wallet](https://internetcomputer.org/docs/building-apps/canister-management/cycles-wallet) with some cycles to pay for requests.
+
+#### Example with [`getSlot`](https://solana.com/de/docs/rpc/http/getslot)
+
+To get the last `finalized` slot on Solana Mainnet
+
+```bash
+dfx canister call --ic sol_rpc --wallet $(dfx identity get-wallet --ic) --with-cycles 2B getSlot \
```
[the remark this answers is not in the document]
gregorydemay: Good idea. Done.

PR #69, opened 2025-04-15
refactor: use SOL RPC canister in `basic_solana` example

lpahlavi:
(XC-345) Until now, the `basic_solana` examples were making RPC requests using manually built HTTP outcalls directly to `api.devnet.solana.com`. This PR refactors the examples to instead call the SOL RPC canister via the `sol_rpc_client`.


examples/basic_solana/src/lib.rs
```diff
@@ -311,31 +315,51 @@ pub async fn send_spl_token(
...
+        .blockhash;
+    Hash::from_str(&blockhash).expect("Unable to parse blockhash")
 }
 
-// TODO: Remove!
-#[query(name = "__transform_json_rpc", hidden = true)]
-fn transform(args: TransformArgs) -> HttpResponse {
-    transform_http_request(args)
+fn client() -> SolRpcClient<IcRuntime> {
+    SolRpcClient::builder_for_ic()
```
gregorydemay, line 358: I think there it would make sense to have the canister ID of the SOL RPC canister configurable so that people could try with different instances (default would be to use `builder_for_ic`). This would also be needed for testing purposes.

examples/basic_solana/src/lib.rs
```diff
@@ -311,31 +315,51 @@ pub async fn send_spl_token(
...
+    client
+        .send_transaction(transaction)
+        .send()
+        .await
+        .expect_consistent()
+        .expect("Call to `sendTransaction` failed")
+        .to_string()
+}
+
+async fn get_recent_blockhash(rpc_client: &SolRpcClient<IcRuntime>) -> Hash {
```
gregorydemay, line 339: I think here it would be nice to have some retry mechanisms (say try at most 3 times) since:

1. `getSlot` could fail (boundary condition)
2. There is not necessarily a block for a slot, so we should not panic

gregorydemay (commented):
Thanks @lpahlavi for the PR! Did you already try out this new version of the examples?

examples/basic_solana/src/lib.rs
lpahlavi: Sure, I think that's a good idea. I think it's still nice to print some information about why the last try failed, so the code is slightly more complicated. Let me know if you have a more elegant way of doing this as a Rust expert!

lpahlavi (commented):
Thanks for the feedback @gregorydemay! I haven't tested the new canister yet, but I will do that this afternoon to make sure everything works as expected.

examples/basic_solana/src/lib.rs
gregorydemay: I think we can do a bit better while returning all errors encountered:
```rust
async fn get_recent_blockhash(rpc_client: &SolRpcClient<IcRuntime>) -> Hash {
    let num_tries = 3;
    let mut errors = Vec::with_capacity(num_tries);
    loop {
        if errors.len() >= num_tries {
            panic!("Failed to get recent block hash after {num_tries} tries: {errors:?}");
        }
        match rpc_client.get_slot().send().await {
            MultiRpcResult::Consistent(Ok(slot)) => match rpc_client.get_block(slot).send().await {
                MultiRpcResult::Consistent(Ok(Some(block))) => {
                    return Hash::from_str(&block.blockhash).expect("Unable to parse blockhash")
                }
                MultiRpcResult::Consistent(Ok(None)) => {
                    errors.push(format!("No block for slot {slot}"));
                    continue;
                }
                MultiRpcResult::Inconsistent(results) => {
                    errors.push(format!(
                        "Inconsistent results for block with slot {slot}: {:?}",
                        results
                    ));
                    continue;
                }
                MultiRpcResult::Consistent(Err(e)) => {
                    errors.push(format!("Failed to get block with slot {slot}: {:?}", e));
                    continue;
                }
            },
            MultiRpcResult::Inconsistent(results) => {
                errors.push(format!("Failed to retrieved last slot: {:?}", results));
                continue;
            }
            MultiRpcResult::Consistent(Err(e)) => {
                errors.push(format!("Failed to retrieve slot: {:?}", e));
                continue;
            }
        }
    }
}
```

gregorydemay (approved):

ninegua (approved):

PR #77, opened 2025-04-23
feat: default commitment level for `SolRpcClient`

gregorydemay:
Follow-up on #75 that improves the `SolRpcClient` by being able to set in the client builder a default commitment level that will be used by the built client for all future requests. This is similar to what the Solana client offers with [`new_with_commitment`](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.new_with_commitment).

### DIFF
--- a/libs/client/src/request/mod.rs
+++ b/libs/client/src/request/mod.rs
@@ -5,9 +5,9 @@ use crate::{Runtime, SolRpcClient};
 use candid::CandidType;
 use serde::de::DeserializeOwned;
 use sol_rpc_types::{
-    AccountInfo, CommitmentLevel, ConfirmedBlock, GetAccountInfoParams, GetBalanceParams,
-    GetBlockParams, GetSlotParams, GetSlotRpcConfig, GetTransactionParams, Lamport, RpcConfig,
-    RpcResult, RpcSources, SendTransactionParams, Signature, TransactionInfo,
+    AccountInfo, ConfirmedBlock, GetAccountInfoParams, GetBalanceParams, GetBlockParams,
+    GetSlotParams, GetSlotRpcConfig, GetTransactionParams, Lamport, RpcConfig, RpcResult,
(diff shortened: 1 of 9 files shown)


libs/client/src/request/mod.rs
```diff
@@ -433,3 +525,9 @@ impl<R: Runtime, Config, Params> RequestCostBuilder<R, Config, Params> {
         self.client.execute_cycles_cost_request(self.request).await
     }
 }
+
+fn set_default<T>(default_value: Option<T>, value: &mut Option<T>) {
```
ninegua, line 529: The current implementation seems to imply that calling set_default the second time will not be effective. Is this desirable?

libs/client/src/request/mod.rs
```diff
@@ -145,8 +153,25 @@ impl SolRpcRequest for GetBlockRequest {
         SolRpcEndpoint::GetBlock
     }
 
-    fn params(self) -> Self::Params {
-        self.0
+    fn params(self, default_commitment_level: Option<CommitmentLevel>) -> Self::Params {
+        let mut params = self.0;
+        let default_block_commitment_level =
+            default_commitment_level
+                .clone()
```
ninegua, line 160: this clone seems redundant?

libs/client/src/request/mod.rs
gregorydemay: I think this is ok in the sense it's meant to be used once. If there is a default value and there is currently no value, then the default value will be used. Calling this method again will have the same result, since now that there is a value, it will not be overriden.

libs/client/src/request/mod.rs
gregorydemay: Good catch, I'm surprised this was not spotted by the linter 🤷‍♂️ .

ninegua (approved):

PR #78, opened 2025-04-23
test: transfer of SPL token in `basic_solana`

gregorydemay:
Follow-up on #75 to also test the transfer of SPL tokens in the `basic_solana` example.

### DIFF
--- a/examples/basic_solana/src/main.rs
+++ b/examples/basic_solana/src/main.rs
@@ -101,7 +101,7 @@ pub async fn get_nonce(account: Option<String>) -> String {
 }
 
 #[update]
-pub async fn get_spl_token_balance(account: Option<String>, mint_account: String) -> String {
+pub async fn get_spl_token_balance(account: Option<String>, mint_account: String) -> Nat {
     let account = account.unwrap_or(associated_token_account(None, mint_account).await);
 
     let commitment = read_state(State::solana_commitment_level);
@@ -138,7 +138,8 @@ pub async fn get_spl_token_balance(account: Option<String>, mint_account: String
     response["value"]["amount"]
         .as_str()
         .unwrap_or_else(|| panic!("Failed to parse getTokenAccountBalance response {response}"))
-        .to_string()
+        .parse()
+        .unwrap()
 }
 
 #[update]
--- a/examples/basic_solana/src/main.rs
+++ b/examples/basic_solana/src/main.rs
@@ -7,7 +7,7 @@ use candid::{Nat, Principal};
 use ic_cdk::{init, post_upgrade, update};
 use num::ToPrimitive;
 use serde_json::json;
-use sol_rpc_types::{CommitmentLevel, GetAccountInfoEncoding, GetAccountInfoParams};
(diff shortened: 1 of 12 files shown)


examples/basic_solana/src/spl.rs
```diff
@@ -4,8 +4,8 @@ use solana_pubkey::Pubkey;
 mod system_program {
     solana_pubkey::declare_id!("11111111111111111111111111111111");
 }
-mod token_program {
-    solana_pubkey::declare_id!("TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");
```
gregorydemay, line 8: This was using the old token program and not the new SPL token 2022 version.

examples/basic_solana/tests/tests.rs
```diff
@@ -243,6 +352,157 @@ impl Setup {
...
+            .get_minimum_balance_for_rent_exemption(MIN_ACCOUNT_LEN as usize)
+            .unwrap();
+        let create_mint_account_ix = solana_program::system_instruction::create_account(
+            &mint_authority.pubkey(),
+            &mint_account.pubkey(),
+            mint_rent,
+            MIN_ACCOUNT_LEN as u64,
+            &SPL_TOKEN_2022_ID,
+        );
+        // See https://github.com/solana-program/token-2022/blob/644f0b014cbdb25c11c20ccedfb6e412d399b6dc/program/src/instruction.rs#L1207
```
gregorydemay, line 372: basically re-do what the spl-token-2022 crate does without pulling the dependency to avoid a hell of dependency conflicts.

examples/basic_solana/src/main.rs
```diff
@@ -95,35 +95,51 @@ pub async fn get_nonce(account: Option<String>) -> String {
...
-        ["result"]["value"]["uiAmountString"]
+    // The response to a json_request for the `getTokenAccountBalance` endpoint has the following format:
+    //{
+    //    "context": {
+    //      "apiVersion": "2.0.22",
+    //      "slot": 63
+    //    },
+    //    "value": {
+    //      "amount": "999999000",
+    //      "decimals": 9,
```
ninegua, line 131: Shouldn't the return value include the decimal too?

ninegua (approved):

examples/basic_solana/src/main.rs
gregorydemay: Do you mean to return the value corresponding to `uiAmount` instead of `amount`? For an example I have the impression that either way is fine. WDYT?

ninegua: Since the SOL RPC returns both `amount` and `decimals`, why not return both instead of just `amount`? Is there another endpoint to lookup `decimals` of a given token? If not, then I feel that this is necessary for the client to know.

gregorydemay: Thanks for the explanation! You're right of course and that will be the future API. Right now this simply serves as an example and should be expanded once we have support for `getTokenAccountBalance` (XC-325). I added a note in the ticket to not forget it.

PR #90, opened 2025-05-05
feat: add support for `getTokenAccountBalance` RPC method

lpahlavi:
(XC-325) Add support for the `getTokenAccountBalance` Solana RPC method.

### DIFF
--- a/canister/sol_rpc_canister.did
+++ b/canister/sol_rpc_canister.did
@@ -579,6 +579,23 @@ type MultiGetSlotResult = variant {
     Inconsistent : vec record { RpcSource; GetSlotResult };
 };
 
+// The parameters for a Solana `getTokenAccountBalance` RPC method call.
+type GetTokenAccountBalanceParams = record {
+  // Pubkey of token account to query, as base-58 encoded string.
+  pubkey: Pubkey;
+  // The commitment describes how finalized a block is at that point in time.
+  commitment: opt CommitmentLevel;
+};
+
+// Represents the result of a call to the `getTokenAccountBalance` Solana RPC method.
+type GetTokenAccountBalanceResult = variant { Ok : TokenAmount; Err : RpcError };
+
+// Represents an aggregated result from multiple RPC calls to the `getTokenAccountBalance` Solana RPC method.
+type MultiGetTokenAccountBalanceResult = variant {
+    Consistent : GetTokenAccountBalanceResult;
+    Inconsistent : vec record { RpcSource; GetTokenAccountBalanceResult };
+};
+
 // Represents the result of a call to the `sendTransaction` Solana RPC method.
 type SendTransactionResult = variant { Ok : Signature; Err : RpcError };
 
@@ -663,11 +680,11 @@ service : (InstallArgs,) -> {
   // The caller is the controller or a principal specified in `InstallArgs::manage_api_keys`.
   updateApiKeys : (vec record { SupportedProvider; opt text }) -> ();
 
-  // Call the Solana `getAccountInfo` RPC method and return the resulting slot.
+  // Call the Solana `getAccountInfo` RPC method and return the resulting info.
   getAccountInfo : (RpcSources, opt RpcConfig, GetAccountInfoParams) -> (MultiGetAccountInfoResult);
   getAccountInfoCyclesCost : (RpcSources, opt RpcConfig, GetAccountInfoParams) -> (RequestCostResult) query;
 
-  // Call the Solana `getBalance` RPC method and return the resulting block.
+  // Call the Solana `getBalance` RPC method and return the resulting balance.
   getBalance : (RpcSources, opt RpcConfig, GetBalanceParams) -> (MultiGetBalanceResult);
   getBalanceCyclesCost : (RpcSources, opt RpcConfig, GetBalanceParams) -> (RequestCostResult) query;
 
@@ -679,7 +696,11 @@ service : (InstallArgs,) -> {
   getSlot : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (MultiGetSlotResult);
   getSlotCyclesCost : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (RequestCostResult) query;
 
-  // Call the Solana `getTransaction` RPC method and return the resulting slot.
+  // Call the Solana `getTokenAccountBalance` RPC method and return the resulting balance.
+  getTokenAccountBalance : (RpcSources, opt RpcConfig, GetTokenAccountBalanceParams) -> (MultiGetTokenAccountBalanceResult);
+  getTokenAccountBalanceCyclesCost : (RpcSources, opt RpcConfig, GetTokenAccountBalanceParams) -> (RequestCostResult) query;
+
+  // Call the Solana `getTransaction` RPC method and return the resulting transaction.
   getTransaction : (RpcSources, opt RpcConfig, GetTransactionParams) -> (MultiGetTransactionResult);
   getTransactionCyclesCost : (RpcSources, opt RpcConfig, GetTransactionParams) -> (RequestCostResult) query;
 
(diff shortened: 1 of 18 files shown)


examples/basic_solana/basic_solana.did
```diff
@@ -79,10 +79,10 @@ service : (InitArg) -> {
     get_nonce : (account: opt Address) -> (Blockhash);
 
     // Returns the balance of the given Solana account for the SPL token associated with
-    // the given token mint account.
+    // the given token mint account formatted as a string.
     //
     // If no account is provided, the account derived from the caller's principal is used.
-    get_spl_token_balance : (account: opt Address, mint_account: Address) -> (nat);
+    get_spl_token_balance : (account: opt Address, mint_account: Address) -> (text);
```
lpahlavi, line 85: I went for the simple option of returning the already formatted balance as I find it adds unnecessary complexity to introduce a new Candid type for `UiTokenAmount` given that this is just a basic example.

integration_tests/tests/solana_test_validator.rs
```diff
@@ -309,6 +309,11 @@ async fn should_get_balance() {
     assert_eq!(compare_balances(&setup, publickey).await, 10_000_000_000);
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn should_get_token_account_balance() {
+    // TODO XC-325: Add test for `getTokenAccountBalance` (requires some SPL test infrastructure)
```
lpahlavi, line 314: I left this out of the PR as it requires a little bit of infrastructure around SPL tokens and ATAs (all of which I should be able to re-use from `basic_solana`, but I would add some of these functionalities to the client, e.g. deriving an ATA) and I didn't want it to get too big. I will add it in a follow-up PR.

canister/src/rpc_client/sol_rpc/mod.rs
```diff
@@ -26,12 +26,14 @@ pub enum ResponseTransform {
     #[n(2)]
     GetBlock,
     #[n(3)]
-    GetSlot(#[n(0)] RoundingError),
-    #[n(4)]
-    GetTransaction,
+    GetSlot(#[n(4)] RoundingError),
```
ninegua, line 29: Why is this change necessary?

canister/sol_rpc_canister.did
```diff
@@ -679,7 +696,11 @@ service : (InstallArgs,) -> {
   getSlot : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (MultiGetSlotResult);
   getSlotCyclesCost : (RpcSources, opt GetSlotRpcConfig, opt GetSlotParams) -> (RequestCostResult) query;
 
-  // Call the Solana `getTransaction` RPC method and return the resulting slot.
+  // Call the Solana `getTokenAccountBalance` RPC method and return the resulting balance.
+  getTokenAccountBalance : (RpcSources, opt RpcConfig, GetTokenAccountBalanceParams) -> (MultiGetTokenAccountBalanceResult);
```
gregorydemay, line 700: nit: we should maybe mention that this will return a JSON-RPC error if the account does not exist. Just tested manually, in that case it returns something like
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid param: not a Token account"
    },
    "id": 1
}
```

canister/src/rpc_client/mod.rs
```diff
@@ -193,6 +193,33 @@ impl GetSlotRequest {
...
+    pub fn get_token_account_balance<Params: Into<json::GetTokenAccountBalanceParams>>(
+        rpc_sources: RpcSources,
+        config: RpcConfig,
+        params: Params,
+    ) -> Result<Self, ProviderError> {
+        let consensus_strategy = config.response_consensus.unwrap_or_default();
+        let providers = Providers::new(rpc_sources, consensus_strategy.clone())?;
+        let max_response_bytes = config
+            .response_size_estimate
+            .unwrap_or(256 + HEADER_SIZE_LIMIT);
```
gregorydemay, line 211: I'm getting 733 Bytes in the case of a standard response, so I would bump that, e.g.
```suggestion
            .unwrap_or(1024 + HEADER_SIZE_LIMIT);
```

canister/src/rpc_client/tests.rs
```diff
@@ -154,6 +153,36 @@ mod request_serialization_tests {
         );
     }
 
+    #[test]
+    fn should_serialize_get_token_account_balance_request() {
+        let pubkey = solana_pubkey::Pubkey::default();
+        assert_serialized(
```
gregorydemay, line 159: nit: maybe rename to something like `assert_params_eq`? I currently find it a bit weird that it takes as 1st parameter the whole request, but only looks at its parameters.

examples/basic_solana/basic_solana.did
gregorydemay: This was mentioned in a previous PR by @ninegua , but we need here also the number of decimals for the example to make sense. Since we can reuse the `sol_rpc_types::TokenAmount`, I think it's just a matter of copying the Candid declaration here, which should not be too much. WDYT?

examples/basic_solana/src/main.rs
```diff
@@ -101,45 +99,16 @@ pub async fn get_nonce(account: Option<String>) -> String {
...
-    //}
-    let response: serde_json::Value = serde_json::from_str(&response)
-        .expect("`getTokenAccountBalance` response is not a valid JSON");
-    response["value"]["amount"]
-        .as_str()
-        .unwrap_or_else(|| panic!("Failed to parse getTokenAccountBalance response {response}"))
-        .parse()
-        .unwrap()
+        .expect("Call to `getTokenAccountBalance` failed")
+        .ui_amount_string
```
gregorydemay, line 111: I would return the full output

examples/basic_solana/tests/tests.rs
```diff
@@ -206,30 +204,30 @@ fn test_basic_solana() {
         .unwrap()
         .expect("Missing user's associated token account");
     assert_eq!(token_account.token_amount.amount, "999999000");
-    let sender_spl_balance: Nat = basic_solana.update_call(
+    let sender_spl_balance: String = basic_solana.update_call(
```
gregorydemay, line 207: It makes more sense to have the balance has an integer IMO, but that should be trivial once we give back the `TokenAmount`.

libs/client/src/lib.rs
```diff
@@ -396,6 +397,61 @@ impl<R> SolRpcClient<R> {
...
+    /// #       ui_amount: None,
+    /// #       decimals: 0,
+    /// #       amount: "".to_string(),
+    /// #       ui_amount_string: "".to_string(),
+    /// #    })))
+    ///     .with_rpc_sources(RpcSources::Default(SolanaCluster::Mainnet))
+    ///     .build();
+    ///
+    /// let balance = client
+    ///     .get_token_account_balance(pubkey!("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v"))
```
gregorydemay, line 424: nit: I know it's just a mock, but this address concretely is invalid (not a Token account)
```suggestion
    ///     .get_token_account_balance(pubkey!("3emsAVdmGKERbHjmGfQ6oZ1e35dkf5iYcS6U4CPKFVaa"))
```

libs/client/src/lib.rs
```diff
@@ -396,6 +397,61 @@ impl<R> SolRpcClient<R> {
...
+    ///     .send()
+    ///     .await
+    ///     .expect_consistent();
+    ///
+    /// assert_eq!(balance, Ok(UiTokenAmount {
+    ///     ui_amount: None,
+    ///     decimals: 0,
+    ///     amount: "".to_string(),
+    ///     ui_amount_string: "".to_string(),
```
gregorydemay, line 433: nit: it would be more useful for the example to actually return something meaningful, e.g. for `3emsAVdmGKERbHjmGfQ6oZ1e35dkf5iYcS6U4CPKFVaa` it could be something like
```suggestion
    ///     ui_amount: Some(251153323.575906),
    ///     decimals: 6,
    ///     amount: "251153323575906".to_string(),
    ///     ui_amount_string: "251153323.575906".to_string(),
```

gregorydemay (approved):
Thanks for this PR @lpahlavi ! Some minor comments.

ninegua (approved):

PR #95, opened 2025-05-07
test: add integration test for `getTokenAccountBalance`

lpahlavi:
(XC-325) Add integration test for `getTokenAccountBalance` with Solana validator.


integration_tests/src/spl.rs
lpahlavi, line 1: This code is copied from `basic_solana`. It seemed easier to me than creating a crate with shared test fixtures. Another alternative would be to add some of this to the client crate as "utils", but since it should hopefully not be necessary anymore once the PR is merged in `anza-xyz/solana-sdk` I think it's an OK temporary solution. WDYT?

integration_tests/tests/solana_test_validator.rs
```diff
@@ -417,6 +476,146 @@ impl Setup {
...
+        self.solana_client
+            .send_transaction(&transaction)
+            .expect("Unable to create associated token account");
+
+        self.wait_for_account_to_exist(&associated_token_account);
+
+        associated_token_account
+    }
+
+    fn create_spl_token(&self) -> (Keypair, Pubkey) {
```
lpahlavi, line 505: Again, this code here is more or less 1-1 copied from the `basic_solana` int tests, but I thought it might be overkill to create a shared test fixtures crate. Feel free to challenge this.

ninegua: Usually copying is fine if I don't expect the code to change much, otherwise it may be worth the effort to refactor.

ninegua (approved):

integration_tests/src/spl.rs
```diff
@@ -0,0 +1,73 @@
...
+    solana_pubkey::declare_id!("ATokenGPvbdGVxr1b2hvZbsiqW5xWH25efTNsLJA8knL");
+}
+
+// TODO XC-297: Replace usages with call to `get_associated_token_address` method in
+//   `spl-associated-token-account-client` crate.
+/// Derives the Associated Token Account address for the given mint address.
+///
+/// This implementation was taken from the `spl-associated-token-account-client` crate
+/// [here](https://github.com/solana-program/associated-token-account/blob/109de0bf04dc033873941c6befed2a7ab07a93d9/interface/src/address.rs#L27)
+/// to force usage of the DFINITY forks for the Solana SDK crates.
```
gregorydemay, line 26: A project that would use SOL RPC to control some SPL tokens from a canister would have the same problem, no? In that sense, having those utility method in a separate crate might be beneficial, WDYT? We could have them for now in a separate crate but not yet published it unless there is demand for it.

integration_tests/tests/solana_test_validator.rs
gregorydemay: similar remark as [here](https://github.com/dfinity/sol-rpc-canister/pull/95/files#r2080956900)

integration_tests/tests/solana_test_validator.rs
```diff
@@ -26,10 +29,13 @@ use solana_transaction_status_client_types::UiTransactionEncoding;
 use std::{
     future::Future,
     str::FromStr,
+    thread,
     thread::sleep,
     time::{Duration, Instant},
 };
 
+pub const SPL_TOKEN_2022_ID: Pubkey = pubkey!("TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb");
```
gregorydemay, line 37: nit: this is already declared in `spl.rs`, no?

integration_tests/tests/solana_test_validator.rs
```diff
@@ -311,7 +317,61 @@ async fn should_get_balance() {
...
+            .compare_client(
+                |sol| {
+                    sol.get_token_account_balance(&account)
+                        .expect("Failed to get token account balance")
+                },
+                |ic| async move {
+                    ic.get_token_account_balance(pubkey)
+                        .modify_params(|params| {
+                            params.commitment = Some(CommitmentLevel::Confirmed)
+                        })
```
gregorydemay, line 332: nit: I have the impression we could simplify all the tests here by changing `icp_client` to
```rust
    fn icp_client(&self) -> SolRpcClient<PocketIcLiveModeRuntime> {
        self.setup
            .client_live_mode()
            .with_default_commitment_level(CommitmentLevel::Confirmed)
            .build()
    }
```
so that the client uses by default `Confirmed` and `modify_params` is then no longer necessary. Better to do that in a separate PR, WDYT?

gregorydemay (approved):
Thanks for this PR @lpahlavi ! Main question for me is whether to put the common code related to SPL in some separate crate if it could be useful to developers using SOL RPC

integration_tests/tests/solana_test_validator.rs
lpahlavi: Agreed, good idea! I will do that in a separate PR.

integration_tests/src/spl.rs
lpahlavi: As discussed offline, we will keep this code until the PR in `agave-xyz/solana-sdk` is merged and the relevant dependencies are updated in the `solana-program/associated-token-account` and `solana-program/token` repositories.

integration_tests/src/spl.rs
lpahlavi: See [this](https://github.com/dfinity/sol-rpc-canister/pull/95/files/eddb9510bf9c2648e6a5f4cedabbb5253e0ce3aa#r2081558213) comment.

integration_tests/tests/solana_test_validator.rs
lpahlavi: See [this](https://github.com/dfinity/sol-rpc-canister/pull/95/files/eddb9510bf9c2648e6a5f4cedabbb5253e0ce3aa#r2081558213) comment.

integration_tests/tests/solana_test_validator.rs
lpahlavi: Good point, but since `spl.rs` is meant to be temporary (and it is defined without `pub` there), I leave it as is for now.

PR #96, opened 2025-05-08
feat: add support for `getSignatureStatuses` RPC method

lpahlavi:
(XC-291) Add support for `getSignatureStatuses` RPC method.

Due to the fast-changing value of the `confirmations` field in the response of `getSignatureStatuses`, it is set to `null` in the response transform. This was done to achieve consensus. Rounding is not performed on this value since a value of `null` is anyways returned after a supermajority of the cluster finalized the transaction, meaning its value is short-lived and the rounded value would not be of much use.


canister/src/rpc_client/json/mod.rs
```diff
@@ -163,6 +163,43 @@ pub struct GetBlockConfig {
...
+impl GetSignatureStatusesParams {
+    pub fn num_signatures(&self) -> usize {
+        self.0.len()
+    }
+}
+
+impl From<sol_rpc_types::GetSignatureStatusesParams> for GetSignatureStatusesParams {
+    fn from(params: sol_rpc_types::GetSignatureStatusesParams) -> Self {
+        let config = if params.is_default_config() {
+            None
```
gregorydemay, line 180: nit: the method `GetSignatureStatusesParams::is_default_config` is I think a bit error-prone. because it ignores the field `signatures` which is because here (in another crate) we split the struct between its 2 fields. I would refactor it so that anything related to that split is local to that file and doesn't transpire somewhere else. Concretely:

1. Remove `GetSignatureStatusesParams::is_default_config` 
2. Move that logic to here, e.g.

```rust
impl From<sol_rpc_types::GetSignatureStatusesParams> for GetSignatureStatusesParams {
    fn from(
        sol_rpc_types::GetSignatureStatusesParams {
            signatures,
            search_transaction_history,
        }: sol_rpc_types::GetSignatureStatusesParams,
    ) -> Self {
        Self(
            signatures,
            search_transaction_history.map(GetSignatureStatusesConfig),
        )
    }
}
```

canister/src/rpc_client/json/mod.rs
```diff
@@ -163,6 +163,43 @@ pub struct GetBlockConfig {
...
+    fn from(params: GetSignatureStatusesParams) -> Self {
+        (params.0, params.1)
+    }
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct GetSignatureStatusesConfig {
+    #[serde(rename = "searchTransactionHistory")]
+    pub search_transaction_history: Option<bool>,
```
gregorydemay, line 200: why the additional option here, given that the top-level type already contains `Option<GetSignatureStatusesConfig>`?

```suggestion
    pub search_transaction_history: bool,
```

canister/sol_rpc_canister.did
```diff
@@ -558,6 +568,41 @@ type TransactionTokenBalance = record {
   uiTokenAmount : TokenAmount;
 };
 
+// Solana transaction status as returned by the `getSignatureStatuses` RPC method.
+//
+// *WARNING*: The optional `confirmations` field in the `getSignatureStatuses` response is not
+//   included in this type since it is ignored when processing the RPC response.
+type TransactionStatus = record {
+    // The slot the transaction was processed.
+    slot: nat64;
```
gregorydemay, line 577: nit: use `Slot` alias type
```suggestion
    slot: Slot;
```

canister/src/rpc_client/mod.rs
```diff
@@ -165,6 +165,34 @@ impl GetBlockRequest {
...
+        rpc_sources: RpcSources,
+        config: RpcConfig,
+        params: Params,
+    ) -> Result<Self, ProviderError> {
+        let params = params.into();
+        let consensus_strategy = config.response_consensus.unwrap_or_default();
+        let providers = Providers::new(rpc_sources, consensus_strategy.clone())?;
+        let max_response_bytes = config
+            .response_size_estimate
+            .unwrap_or(128 + (params.num_signatures() as u64 * 256) + HEADER_SIZE_LIMIT);
```
gregorydemay, line 184: nice! Maybe a comment explaining how this heuristic was computed would be nice.

canister/src/rpc_client/sol_rpc/mod.rs
```diff
@@ -26,22 +27,36 @@ pub enum ResponseTransform {
...
     #[n(7)]
+    SendTransaction,
+    #[n(8)]
     Raw,
 }
 
 impl ResponseTransform {
     fn apply(&self, body_bytes: &mut Vec<u8>) {
+        #[derive(Clone, Debug, Deserialize, Serialize)]
+        pub struct SolanaRpcResult<T> {
```
gregorydemay, line 46: nice!

canister/src/rpc_client/tests.rs
```diff
@@ -82,6 +83,34 @@ mod request_serialization_tests {
...
+                RpcSources::Default(SolanaCluster::Mainnet),
+                RpcConfig::default(),
+                GetSignatureStatusesParams {
+                    signatures: vec![],
+                    search_transaction_history: None,
+                },
+            )
+            .unwrap(),
+            json!([[], null]),
+        );
```
gregorydemay, line 111: Isn't it the same thing as in L.88?

libs/client/src/lib.rs
```diff
@@ -452,6 +454,86 @@ impl<R> SolRpcClient<R> {
...
+    ///         confirmation_status: Some(TransactionConfirmationStatus::Finalized),
+    ///     }),
+    /// ]));
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[allow(clippy::type_complexity)]
+    pub fn get_signature_statuses(
+        &self,
+        params: impl Into<GetSignatureStatusesParams>,
```
gregorydemay, line 520: nit: We could provide a slightly better API by abstracting away the exact collection type used to be able to write something like
```rust
client.get_signature_status[sig_1, sig_2]
  .with_search_transaction_history(true)
```
or
```rust
client.get_signature_status[vec![sig_1, sig_2]]
  .with_search_transaction_history(true)
```
or
```rust
client.get_signature_status[btreeset!{sig_1, sig_2}]
  .with_search_transaction_history(true)
```
For this, you could use the `IntoIterator` trait like in [here](https://github.com/dfinity/sol-rpc-canister/blob/4df3475f41f9f9b5efafe95e54a97a3d36f4a9f6/libs/client/src/lib.rs#L523) and add some dedicated method `with_search_transaction_history`, similar to what was done [here](https://github.com/dfinity/sol-rpc-canister/blob/4df3475f41f9f9b5efafe95e54a97a3d36f4a9f6/libs/client/src/request/mod.rs#L482)

gregorydemay (commented):
Thanks for this PR @lpahlavi! A few comments, mostly minor.

canister/sol_rpc_canister.did
lpahlavi: Good catch, thanks! Done.

canister/sol_rpc_canister.did
```diff
@@ -558,6 +568,41 @@ type TransactionTokenBalance = record {
   uiTokenAmount : TokenAmount;
 };
 
+// Solana transaction status as returned by the `getSignatureStatuses` RPC method.
+//
+// *WARNING*: The optional `confirmations` field in the `getSignatureStatuses` response is not
+//   included in this type since it is ignored when processing the RPC response.
```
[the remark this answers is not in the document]
lpahlavi: Good idea. Done.

canister/src/rpc_client/json/mod.rs
lpahlavi: Hm, I'm not quite sure I follow. This is the exact same as all the other `is_default_config` methods, isn't it? Do you find that these are all error-prone? IMO this just defines which fields are optional config fields, and which ones are required, which is fine where the types are defined. Or maybe I am missing something?

canister/src/rpc_client/json/mod.rs
lpahlavi: This has the slight disadvantage of us having to define a default value for this config field. Though this is only theoretical since in practice the field should always be `Some(..)` if the config is not default.

integration_tests/tests/solana_test_validator.rs
```diff
@@ -307,13 +308,65 @@ async fn should_get_balance() {
...
+        setup
+            .solana_client
+            .request_airdrop(&Keypair::new().pubkey(), 10_000_000_000)
+            .expect("Error while requesting airdrop"),
+        // An arbitrary signature not corresponding to any transaction
+        Signature::from([57u8; 64]),
+    ];
+    let signatures_copy = signatures.clone();
+
+    setup.confirm_transaction(&signatures[0]);
```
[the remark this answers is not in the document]
lpahlavi: Thanks for the suggestion! Done.

libs/types/src/solana/request/mod.rs
```diff
@@ -173,6 +173,39 @@ pub enum TransactionDetails {
     None,
 }
 
+/// The parameters for a Solana [`getSignatureStatuses`](https://solana.com/docs/rpc/http/getsignaturestatuses) RPC method call.
+#[derive(Debug, Clone, Default, Deserialize, Serialize, CandidType)]
+pub struct GetSignatureStatusesParams {
+    /// An array of transaction signatures to confirm, as base-58 encoded strings (up to a maximum of 256)
+    pub signatures: Vec<Signature>,
```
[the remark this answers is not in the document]
lpahlavi: Good idea (I used `Borrow<Signature>` so this works with both `Signature` and `&Signature` though)! I'm not sure I understand the point of the `serde` `try_from` annotation here actually. Otherwise, done.

canister/src/rpc_client/tests.rs
lpahlavi: Good catch, I forgot to populate the second test! Done.

libs/client/src/lib.rs
lpahlavi: Good idea! Done.

canister/src/rpc_client/mod.rs
lpahlavi: I just did this experimentally with `curl`. Do you think this is worth mentioning? IMO this is pretty much the same as all other methods, no?

libs/types/src/solana/request/mod.rs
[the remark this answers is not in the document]
gregorydemay: > I'm not sure I understand the point of the serde try_from annotation here actually.

This is to ensure that when the user inputs will be deserialized by Candid (that actually uses serde), deserialization will fail if there are too many elements. Otherwise, you could still create instances of `GetSignatureStatusesParams` which do not satisfy this invariant. I don't know whether the `try_from` attribute will accept a generic implementation, so you may need to refactor this.

canister/src/rpc_client/json/mod.rs
gregorydemay: > This is the exact same as all the other is_default_config methods, isn't it? Do you find that these are all error-prone? 

Good point, I agree that it is similar to the offer `is_default_config` methods. Looking back I think I would inlined all those methods (separate PR):

1. They are now public methods in the `sol_rpc_types`
2. They are only used internally to transform the Candid type into some JSON type and the notion of what is a default field only makes sense in that context.
 
In other words, this is a leaky abstraction and that's why I would not expose those methods. Does it make sense?

For this concrete new type `GetSignatureStatusesParams` I think the implementation I suggested seems to me much simpler than going through some intermediate method. WDYT?

canister/src/rpc_client/json/mod.rs
gregorydemay: > This has the slight disadvantage of us having to define a default value for this config field.

I don't get it, you could just do `search_transaction_history.map(GetSignatureStatusesConfig)` like the suggestion in this [comment](https://github.com/dfinity/sol-rpc-canister/pull/96#discussion_r2084768361)?

canister/src/rpc_client/mod.rs
gregorydemay: I was more wondering how the coefficients `128` and `256` were estimated.

libs/client/src/lib.rs
```diff
@@ -452,6 +454,86 @@ impl<R> SolRpcClient<R> {
...
+        GetSignatureStatusesParams,
+        sol_rpc_types::MultiRpcResult<Vec<Option<TransactionStatus>>>,
+        sol_rpc_types::MultiRpcResult<
+            Vec<Option<solana_transaction_status_client_types::TransactionStatus>>,
+        >,
+    > {
+        RequestBuilder::new(
+            self.clone(),
+            GetSignatureStatusesRequest::new(params.into()),
+            10_000_000_000,
```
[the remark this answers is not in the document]
gregorydemay: > Maybe it would make sense to create a separate ticket

Yes, we actually already have a ticket for this, which is XC-338. Note that you added a todo for XC-339 which is for e2e test for sending transaction so I don't think it fits.

canister/src/rpc_client/json/mod.rs
lpahlavi: > In other words, this is a leaky abstraction and that's why I would not expose those methods. Does it make sense?

I see, what you say makes sense. I guess to me, it also makes sense to define what fields are part of the optional config where the request parameter types themselves are defined, because in a way it belongs with the type definition. On the other hand, it is a little bit of unnecessary bloat in the `sol_rpc_types` crate which is really only used in one spot and could be inlined there. I think this belongs to a separate PR though as you suggested.

> For this concrete new type GetSignatureStatusesParams I think the implementation I suggested seems to me much simpler than going through some intermediate method. WDYT?

Yes, I agree with this type it is much cleaner to skip the intermediate method. Done.

canister/src/rpc_client/json/mod.rs
lpahlavi: You are right yes, this is actually cleaner. Done.

README.md
```diff
@@ -117,6 +117,8 @@ The SOL RPC canister reaches the Solana JSON-RPC providers using [HTTPS outcalls
    1. Use a [durable nonce](https://solana.com/de/developers/guides/advanced/introduction-to-durable-nonces) instead of a blockhash.
    2. Retrieve a recent blockhash by first retrieving a recent slot with `getSlot` and then getting the block (which includes the blockhash) with `getBlock`.
 
+[//]: # (TODO: XC-339: mention also `getSignatureStatuses#confirmations`)
```
gregorydemay, line 120: I don't get why this is referring XC-339? Shouldn't it be this ticket, XC-291?

gregorydemay (commented):
Thanks for the changes! The "main" last point is the deserialization of `GetSignatureStatusesParams` to ensure that it fails if there are too many signatures.

libs/types/src/solana/request/mod.rs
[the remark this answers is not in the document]
lpahlavi: Hm I see that makes sense, I didn't think about the fact that deserialization "gets around" the length limitation. I still don't understand how the Candid type, including the `searchTransactionHistory` field can be deserialized with just `#[serde(try_from = "Vec<solana_signature::Signature>")]`, especially since in the Candid type the signatures are just strings. I'm thinking that the better way to go forward might be to introduce a type for size-limited vectors, similar to e.g. in some of the crypto crates. WDYT?

README.md
lpahlavi: Yes! That's a mistake. Fixed.

canister/src/rpc_client/mod.rs
lpahlavi: Yes, I did this experimentally with `curl` as well (which AFAIK is pretty much how we estimated the heuristics for other methods). The only difference with this endpoint is I tried requests with 0 and `n` signatures in the input, observed the size of the response for each one, and fit a line to the data points (approximately).

libs/client/src/lib.rs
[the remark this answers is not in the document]
lpahlavi: Ah good catch, thanks! Done.

lpahlavi:
Thanks for the re-review @gregorydemay! I added a new type in `sol_rpc_types` to handle vectors with a maximum capacity. I think this makes the most sense IMO since this way once can never instantiate e.g. `getSignatureStatuses` request parameters with too many signatures, whether through deserialization or direct instantiation. As a bonus I could re-use this for the `getRecentPrioritizationFees` request parameters. Let me know what you think!

libs/types/src/solana/request/mod.rs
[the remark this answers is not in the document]
gregorydemay: > I still don't understand how the Candid type, including the searchTransactionHistory field can be deserialized with just #[serde(try_from = "Vec<solana_signature::Signature>")], especially since in the Candid type the signatures are just strings.

Candid for decoding delegates to `serde`. That's why you can use this `try_from` pattern to ensure that only correct data will result from the deserialization. Since in that case `GetSignatureStatusesParams` wraps a bit more than a vec (the field `search_transaction_history`), you could do this kind of pattern as follows

1. Private unchecked type, like `GetSignatureStatusesParamsUnchecked` having the same fields as `GetSignatureStatusesParams`. 
2. Do a `TryFrom` conversion:  to obtain `GetSignatureStatusesParams` from `GetSignatureStatusesParamsUnchecked`

Like this the rest of the codebase only works with `GetSignatureStatusesParams` which is always valid by construction.

>  I'm thinking that the better way to go forward might be to introduce a type for size-limited vectors, similar to e.g. in some of the crypto crates. WDYT?

That's also a valid way forward!

libs/types/src/solana/mod.rs
```diff
@@ -95,3 +96,28 @@ pub struct PrioritizationFee {
     #[serde(rename = "prioritizationFee")]
     pub prioritization_fee: MicroLamport,
 }
+
+/// A vector with a maximum capacity.
+#[derive(Debug, Clone, Deserialize, Serialize, CandidType, PartialEq, Default)]
+#[serde(try_from = "Vec<T>")]
```
gregorydemay, line 102: shouldn't we have also `into` `Vec<T>` for serialization?

```suggestion
#[serde(try_from = "Vec<T>", into="Vec<T>")]
```

libs/types/src/solana/mod.rs
```diff
@@ -95,3 +96,28 @@ pub struct PrioritizationFee {
...
+pub struct VecWithMaxLen<T, const CAPACITY: usize>(Vec<T>);
```
gregorydemay, line 103: Shouldn't that type be outside of `libs/types/src/solana` given that it has nothing to do with Solana? Maybe in `lib.rs` or `rpc_client/`?

libs/types/src/solana/request/tests.rs
```diff
@@ -0,0 +1,62 @@
+use crate::{GetRecentPrioritizationFeesParams, GetSignatureStatusesParams};
+use serde::Deserialize;
+use serde_json::json;
+
+mod get_signature_statuses_params_tests {
```
gregorydemay, line 5: nit: I think you could unify both modules since it seems they rather test the behavior of `VecWithMaxLen` and not the specificities `GetSignatureStatusesParams`/`GetRecentPrioritizationFeesParams`

gregorydemay (approved):
Thanks @lpahlavi for this PR! A couple of minor comments but otherwise LGTM!

lpahlavi (commented):
Thank you for the review @gregorydemay!

libs/types/src/solana/mod.rs
gregorydemay: > Hm I added some tests to make sure that encoding/decoding round trips with Candid work and it seems this is not necessary.

I think that's expected because we only use those Candid types to deserialize user inputs, but the Candid encoding of MaxVecLength<T> would result in different bytes than the Candid encoding of the underlying Vec<T>, which is I think unexpected. To test this you do a test similar to this [one](https://github.com/dfinity/sol-rpc-canister/blob/1744026c3ab9ddd5613ecece68b6e2259946b0ee/libs/types/src/solana/tests.rs#L63).

> adding this forces us to add the Clone restriction on the generic type T

I'm not following why `Clone`, the bound should be `T: Serialize`, no?

PR #97, opened 2025-05-09
test: use default commitment for client in `solana_test_validator.rs`

lpahlavi:
(XC-291) Leverage `ClientBuilder::with_default_commitment` to avoid having to specify the commitment level in all request parameters in `solana_test_validator.rs`.


gregorydemay (approved):
Thanks for this simplification @lpahlavi !

PR #104, opened 2025-05-13
test: end-to-end tests for `sendTransaction` 

lpahlavi:


integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
+
+// This test should be run together with end-to-end tests, not other integration tests
+#[ignore]
+#[tokio::test(flavor = "multi_thread")]
+async fn should_send_transaction() {
+    let setup = Setup::new();
+
+    fn load_keypair(key: &str) -> Keypair {
+        fn try_load_keypair(key: &str) -> Result<Keypair, String> {
```
ninegua, line 28: nit: this inner function seems only here to propagate errors. I think you can just change the test function itself to Result type and use `try_load_keypair(...)?` directly in the test body.

ninegua (dismissed):

.github/workflows/ci.yml
```diff
@@ -175,3 +174,10 @@ jobs:
       - name: "Detect Inconsistent Results"
         working-directory: canister/ci
         run: cat e2e_examples.log | grep -q -e Inconsistent && exit 1 || exit 0
+
+      - name: "Run end-to-end integration tests"
```
gregorydemay, line 178: renaming?

integration_tests/src/wallet.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl TryFrom<CallCanisterArgs> for Vec<u8> {
+    type Error = candid::Error;
+
+    fn try_from(args: CallCanisterArgs) -> Result<Vec<u8>, Self::Error> {
+        Encode!(&args)
+    }
+}
```
gregorydemay, line 42: nit: this I wouldn't do. `From`/`TryFrom` conversion are needed when there is only 1 clear way how a type should be transformed into another type. However, here, there are many ways how `CallCanisterArgs` could be serialized to a blob (Candid, Json, Cbor, etc.)

integration_tests/src/wallet.rs
```diff
@@ -0,0 +1,72 @@
+//! Module to interact with a [cycles wallet](https://github.com/dfinity/cycles-wallet) canister.
```
gregorydemay, line 1: I think this refactoring makes sense, but I would have done it in a separate PR. As far as I can tell it does not have anything to do with the purpose of this PR.

integration_tests/src/lib.rs
```diff
@@ -497,18 +444,76 @@ impl Runtime for PocketIcLiveModeRuntime<'_> {
...
+            .env
+            .query_call(id, self.caller, method, encode_args(args))
+            .await
+            .map_err(PocketIcRuntime::parse_reject_response);
+        decode_call_response(result?)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct IcAgentRuntime<'a> {
```
gregorydemay, line 481: nit: why not moving this struct and its implementation of `runtime` in its own module, e.g. `agent.rs` to be similar to `wallet.rs`? This may be irrelevant if we do the suggestion from this [comment](https://github.com/dfinity/sol-rpc-canister/pull/104#discussion_r2091190865).

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+            let statuses = self
+                .client()
+                .get_signature_statuses([transaction_id])
+                .unwrap()
+                .send()
+                .await
+                .expect_consistent()
+                .unwrap_or_else(|_| {
+                    panic!("Failed to get status for transaction {transaction_id}")
+                });
```
gregorydemay, line 160: nit: since we are trying several times, we should maybe avoid panicking to make the test more robust?

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+    assert!(sender_balance_after + transaction_amount <= sender_balance_before);
+}
+
+pub struct Setup {
+    agent: Agent,
+    sol_rpc_canister_id: CanisterId,
+    wallet_canister_id: CanisterId,
+}
+
+impl Setup {
```
gregorydemay, line 99: Since this uses a completely different setup than the one in `integration_tests/src/lib.rs` maybe it would be better to have end-to-end tests separated from `integration_tests`, meaning in another crate. Also because maybe sometimes we want to run the integration tests, but not necessarily the end-to-end tests. WDYT?

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+        .get_block(slot)
+        .send()
+        .await
+        .expect_consistent()
+        .expect("Call to get block failed")
+        .expect("Block not found");
+    let blockhash = Hash::from_str(&block.blockhash).expect("Failed to parse blockhash");
+
+    let transaction_amount = 1_000;
+    let transaction = Transaction::new_signed_with_payer(
```
gregorydemay, line 60: I think it would be useful to also call `getRecentPrioritizationFees` and use this to set a priority fee+compute unit limit (like in [here](https://github.com/dfinity/sol-rpc-canister/blob/0ffba32c0c08e23c3d55f95f57664f52a8fbd579/integration_tests/tests/solana_test_validator.rs#L85). Based on the [transaction](https://explorer.solana.com/tx/5vxQcYXWDjU8Ao95vRFiFjHjp4zm5krwHC9rTNymqpwgtsC8LzUBjVHTN6RvYWjPb4pwgECk1fzQcTe8xvegfRwh?cluster=devnet) you mentioned, 150 CUL should be ok). This is important because probably needed on mainnet, and since this adds again some latency, we need to make sure that everything works together.

integration_tests/src/wallet.rs
lpahlavi: Yes you are right. I did have to refactor the code in this module to some extent to work with the new `ic-agent` runtime but the changes were minimal.

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+use solana_pubkey::Pubkey;
+use solana_signature::Signature;
+use solana_signer::Signer;
+use solana_transaction::Transaction;
+use std::{env, str::FromStr, time::Duration};
+
+const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
+
+// This test should be run together with end-to-end tests, not other integration tests
+#[ignore]
```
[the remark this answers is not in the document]
lpahlavi: Actually this was a bit of a hack to be able to run this test separately from the other integration tests (to run this test I could specify the test file and add the `--include-ignored` Cargo flag). Definitely not ideal though... Since I moved everything to a dedicated crate this is not needed anymore though!

integration_tests/tests/end_to_end.rs
lpahlavi: Good idea! I added your suggestion. Question: I used the 'max' value from the recent prioritization fees. We could also used e.g. the mean, median, etc. Any preferences? On Devnet this shouldn't really make a difference but we might want to move this test to Mainnet at some point.

integration_tests/tests/end_to_end.rs
lpahlavi: I agree actually. I thought about doing this first, then for some reason thought maybe it was overkill, but upon reflection I think it does make sense. Done.

integration_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    async fn airdrop(&self, account: &Pubkey, amount: u64) -> u64 {
+        let balance_before = self.get_account_balance(account).await;
+        let _airdrop_tx = self
+            .client()
+            .json_request(json!({
+            "jsonrpc": "2.0",
+            "id": 1,
+            "method": "requestAirdrop",
```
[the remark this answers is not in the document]
lpahlavi: Good idea! I think it's definitely not super high priority but would probably be useful at some point. I created a ticket here: [XC-366](https://dfinity.atlassian.net/browse/XC-366).

integration_tests/tests/end_to_end.rs
lpahlavi: Good idea! Done.

integration_tests/tests/end_to_end.rs
lpahlavi: Interesting suggestion! I forgot Rust tests are allowed to return a value. However, I think I prefer to leave it as is because the original error message is quite unhelpful and this way it is clear what the problem is if this ever happens again. WDYT?

lpahlavi (commented):
Thanks a lot for the review feedback @ninegua and @gregorydemay! I've addressed all of your comments. The new diff is on the larger side since I moved the test to a separate crate, however the changes are not as big as it seems.

end_to_end_tests/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
```
gregorydemay, line 1: nit: here for a lack of a better place, could you add the following entry in `release-plz.toml`
```
[[package]]
name = "sol_rpc_e2e_tests"
release = false # don't process this package
```

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,97 @@
...
+    let sender = load_keypair("SOLANA_SENDER_PRIVATE_KEY_BYTES");
+    let recipient = load_keypair("SOLANA_RECEIVER_PRIVATE_KEY_BYTES");
+
+    let sender_balance_before = setup.fund_account(&sender.pubkey(), 1_000_000_000).await;
+    let recipient_balance_before = setup.fund_account(&recipient.pubkey(), 1_000_000_000).await;
+
+    // Set a compute unit (CU) price fee based on the recent prioritization fees
+    let prioritization_fees: Vec<_> = setup
+        .client()
+        .get_recent_prioritization_fees(&[])
```
gregorydemay, line 32: The arguments of `get_recent_prioritization_fees` should contain the addresses of all writable accounts, i.e.  of the sender and the recipient.

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -0,0 +1,97 @@
...
+        .expect("Call to `getRecentPrioritizationFees` failed")
+        .into_iter()
+        .map(|fee| fee.prioritization_fee)
+        .collect();
+    let priority_fee = prioritization_fees.into_iter().max().unwrap_or_default();
+    let add_priority_fee_ix = ComputeBudgetInstruction::set_compute_unit_price(priority_fee);
+
+    // Set a CU limit based for a simple transfer
+    // TODO XC-339: Use reasonable value here
+    let set_cu_limit_ix = ComputeBudgetInstruction::set_compute_unit_limit(1000);
```
gregorydemay, line 46: nit: is that something you want to do in a follow-up PR?

integration_tests/tests/end_to_end.rs
gregorydemay: > I used the 'max' value from the recent prioritization fees. We could also used e.g. the mean, median, etc. Any preferences?  

I think for now this is fine. There is probably a library somewhere to handle this type of calculation. For Mainnet the max is for sure not recommended, there I would rather use something like median.

integration_tests/tests/end_to_end.rs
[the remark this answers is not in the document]
gregorydemay: Thanks for creating the ticket!

gregorydemay (approved):
Thanks @lpahlavi for those e2e tests! having them in a dedicated crate definitely looks cleaner!

end_to_end_tests/Cargo.toml
lpahlavi: Ah yes! Done.

end_to_end_tests/tests/end_to_end.rs
lpahlavi: Good point. Done.

end_to_end_tests/tests/end_to_end.rs
lpahlavi: Actually I just wanted to wait for the tests to go through once (hence the high temporary value) to get an idea of a reasonable value. Done.

integration_tests/tests/end_to_end.rs
lpahlavi: I didn't see any library for this with a quick search, but I did change it to use the median which should be a bit more reasonable but still robust. We can always revisit this later if we have problems when trying to run the tests on Mainnet.

PR #106, opened 2025-05-15
feat: add support for `getSignaturesForAddress`

lpahlavi:
(XC-290) Add support for the `getSignaturesForAddress` Solana RPC method.


libs/client/src/lib.rs
```diff
@@ -530,6 +532,66 @@ impl<R> SolRpcClient<R> {
...
+    ///     .await
+    ///     .expect_consistent();
+    ///
+    /// assert_eq!(statuses, Ok(vec![
+    ///     ConfirmedTransactionStatusWithSignature {
+    ///         signature: Signature::from_str("5h6xBEauJ3PK6SWCZ1PGjBvj8vDdWG3KpwATGy1ARAXFSDwt8GFXM7W5Ncn16wmqokgpiKRLuS83KUxyZyv2sUYv").unwrap(),
+    ///         slot: 114,
+    ///         err: None,
+    ///         memo: None,
+    ///         block_time: None,
```
gregorydemay, line 577: nit: even finalized transactions have a `block_time` I think.

integration_tests/tests/tests.rs
```diff
@@ -1939,6 +1954,82 @@ mod get_signature_statuses_tests {
...
+                "params": [
+                    USDC_PUBLIC_KEY.to_string(),
+                    {
+                        "limit": 1,
+                    },
+                ],
+            })
+        }
+
+        fn response_body(id: u8) -> serde_json::Value {
```
gregorydemay, line 1976: nit: for this kind of tests it's usually better to use a real-world response (to potentially detect some unforeseen edge-cases)

integration_tests/tests/solana_test_validator.rs
```diff
@@ -506,6 +509,79 @@ async fn should_get_signature_statuses() {
...
+        .request_airdrop(&Keypair::new().pubkey(), 10_000_000_000)
+        .expect("Error while requesting airdrop");
+    setup.confirm_transaction(&before);
+
+    let (sol_res, ic_res) = setup
+        .compare_client(
+            |sol| {
+                sol.get_signatures_for_address_with_config(
+                    &Pubkey::default(),
+                    GetConfirmedSignaturesForAddress2Config {
```
gregorydemay, line 533: haha what a name 😃

integration_tests/tests/solana_test_validator.rs
```diff
@@ -506,6 +509,79 @@ async fn should_get_signature_statuses() {
...
+#[tokio::test(flavor = "multi_thread")]
+async fn should_get_signatures_for_address() {
+    let setup = Setup::new().await;
+
+    // Start searching backwards from this transaction. This ensures the two clients get the
+    // same answer. Otherwise, they might get different results due to one of them including
+    // transactions from more recent blocks than others.
+    // Also wait until this transaction is finalized, so that all transactions before it should
+    // be finalized as well. Otherwise, the confirmation status of some transactions might change
+    // while fetching them and the two clients might get different results.
```
gregorydemay, line 521: I think those observations are critical for the correct usage of that method in the context of HTTPs outcalls so we should put it a bit more prominently (Limitations in the readme and also in the Candid file?)

integration_tests/tests/solana_test_validator.rs
```diff
@@ -506,6 +509,79 @@ async fn should_get_signature_statuses() {
...
+        .solana_client
+        .request_airdrop(&Keypair::new().pubkey(), 10_000_000_000)
+        .expect("Error while requesting airdrop");
+    setup.confirm_transaction(&before);
+
+    let (sol_res, ic_res) = setup
+        .compare_client(
+            |sol| {
+                sol.get_signatures_for_address_with_config(
+                    &Pubkey::default(),
```
gregorydemay, line 532: Shouldn't we create some transactions beforehand and use the receiver's or sender's public key as the target of `getSignaturesForAddress`? Otherwise the result here is always empty, no?

gregorydemay (commented):
Thanks @lpahlavi for that PR and sorry for the delay in reviewing it. The code looks good, only some minor comments. The main question is how we could use that method to scrape the transactions of an account programmatically and this is currently less clear to me.

canister/sol_rpc_canister.did
```diff
@@ -718,6 +746,23 @@ type CommitmentLevel = variant {
...
+    // The account address.
+    pubkey: Pubkey;
+    // The commitment describes how finalized a block is at that point in time.
+    commitment: opt CommitmentLevel;
+    // The minimum slot that the request can be evaluated at.
+    minContextSlot: opt Slot;
+    // Maximum transaction signatures to return (between 1 and 1,000).
+    limit: opt nat16;
+    // Start searching backwards from this transaction signature. If not provided the search
+    // starts from the top of the highest max confirmed block.
```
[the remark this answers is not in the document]
lpahlavi: Yes, I guess I am generally weary of adding too much extra documentation to types which are defined by the Solana RPC API, but you're right that in this case some extra IC-specific notes do make sense. PTAL.

In terms of scraping transactions for an account `a` I see things like this:
1. get a recent finalized transaction `t` (e.g. with `getSlot` + `getBlock`)
2. fetch a batch of transactions for `a` searching backwards from `t`, and search no further than the last scraped transaction for `a`
3. keep fetching batches starting from the earliest transaction in the last batch, until the size of the batch is less than `limit`

This should work fine, don't you think?

integration_tests/tests/solana_test_validator.rs
lpahlavi: Good idea. I added a placeholder in the readme and some notes in the Candid file.

integration_tests/tests/solana_test_validator.rs
```diff
@@ -506,6 +509,79 @@ async fn should_get_signature_statuses() {
...
 }
 
+#[tokio::test(flavor = "multi_thread")]
+async fn should_get_signatures_for_address() {
+    let setup = Setup::new().await;
+
+    // Start searching backwards from this transaction. This ensures the two clients get the
+    // same answer. Otherwise, they might get different results due to one of them including
+    // transactions from more recent blocks than others.
+    // Also wait until this transaction is finalized, so that all transactions before it should
```
lpahlavi, line 519: @gregorydemay here, we could also consider ignoring the transaction status of returned transactions, since otherwise we will virtually only be able to reach consensus when all returned transactions are finalized. WDYT?

lpahlavi (commented):
Thanks a lot for your review @gregorydemay! I've addressed/answered all of your comments, PTAL and let me know if you agree.

canister/sol_rpc_canister.did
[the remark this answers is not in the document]
gregorydemay: > Yes, I guess I am generally weary of adding too much extra documentation to types which are defined by the Solana RPC API, but you're right that in this case some extra IC-specific notes do make sense. PTAL.

Thanks for adding the doc, LGTM!

>This should work fine, don't you think?

Yes! I also don't see another way than getting a finalized transaction to bootstrap and potentially needed to fetch many batches of transactions if a lot of transactions are happening on the targeted account. I think we should store this information somewhere, how about adding this high-level algorithm in the design doc (section [Retrieve all transactions](https://docs.google.com/document/d/1H7ysdouS8Owxxe4IUeNJgqR3nrAl3AowXOpT24oq0Fg/edit?tab=t.0#heading=h.g4r65sv7q1o5))?

integration_tests/tests/tests.rs
```diff
@@ -1939,6 +1954,82 @@ mod get_signature_statuses_tests {
...
+                    MockOutcallBuilder::new(200, response_body(first_id + 1))
+                        .with_request_body(request_body(first_id + 1)),
+                    MockOutcallBuilder::new(200, response_body(first_id + 2))
+                        .with_request_body(request_body(first_id + 2)),
+                ])
+                .build()
+                .get_signatures_for_address(USDC_PUBLIC_KEY)
+                .modify_params(|params| {
+                    params.limit = Some(GetSignaturesForAddressLimit::try_from(1).unwrap())
+                })
```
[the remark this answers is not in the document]
gregorydemay: > At least the ones that are somewhat likely to be used, e.g. the ones we actually set in int tests.

Yes currently, that's also my compass. I agree that it may not be super consistent but the main ones should be there. We can always add more if there is demand for it.

integration_tests/tests/solana_test_validator.rs
gregorydemay: > Pubkey::default() is the address of the Solana system program (i.e. 111...111)

good point, I missed that!

gregorydemay (approved):
Thanks @lpahlavi for all the improvements and explanations!

PR #108, opened 2025-05-15
feat: Candid NonZeroU8

gregorydemay:


libs/types/src/rpc_client/mod.rs
```diff
@@ -206,16 +208,30 @@ pub struct GetRecentPrioritizationFeesRpcConfig {
...
     pub max_slot_rounding_error: Option<RoundingError>,
 
-    /// Limit the number of returned priority fees.
+    #[serde(rename = "maxLength")]
+    max_length: Option<NonZeroU8>,
+}
+
+impl GetRecentPrioritizationFeesRpcConfig {
+    /// Default number of priority fees to return.
+    pub const DEFAULT_MAX_LENGTH: NonZeroU8 =
```
lpahlavi, line 217: I'm wondering if it wouldn't make more sense to define a custom type (e.g. `struct MaxLength(NonZeroU8)`) here instead, that then implements the `Default` trait. It's not a huge difference so up to you.

canister/src/rpc_client/mod.rs
lpahlavi, line 1: Here for lack of better place: [here](https://github.com/dfinity/sol-rpc-canister/blob/ed4250bdb24402d8629f1b4cae345d0f5cfe4c86/canister/src/rpc_client/sol_rpc/mod.rs#L111), you can simplify

```rust
if fees.is_empty() || max_length == &0 {
```

to

```rust
if fees.is_empty() {
```

lpahlavi (commented):
Thanks for the PR @gregorydemay! Just 1-2 minor nits from me, otherwise LGTM 🎉

libs/types/src/rpc_client/mod.rs
lpahlavi: Makes sense, thanks for the explanation!

canister/src/rpc_client/mod.rs
lpahlavi: Indeed not as easy as it looks 🙈 Thanks for the very thorough explanation and changes!

lpahlavi (approved):

PR #109, opened 2025-05-16
refactor: Serialize `VecWithMaxLen` properly

lpahlavi:
(XC-291) Serialize the new `VecWithMaxLen` type to `Vec` (i.e. add `#[serde(into = "Vec<T>")]` annotation to type) and add some unit tests to ensure encoding/decoding round trips with Candid are successful when expected or fail when the underlying data is too long.

### DIFF
--- a/libs/types/src/lib.rs
+++ b/libs/types/src/lib.rs
@@ -44,10 +44,10 @@ pub use solana::{
 
 /// A vector with a maximum capacity.
 #[derive(Debug, Clone, Deserialize, Serialize, CandidType, PartialEq, Default, Into)]
-#[serde(try_from = "Vec<T>")]
-pub struct VecWithMaxLen<T, const CAPACITY: usize>(Vec<T>);
+#[serde(try_from = "Vec<T>", into = "Vec<T>")]
+pub struct VecWithMaxLen<T: Clone, const CAPACITY: usize>(Vec<T>);
 
-impl<T, const CAPACITY: usize> TryFrom<Vec<T>> for VecWithMaxLen<T, CAPACITY> {
+impl<T: Clone, const CAPACITY: usize> TryFrom<Vec<T>> for VecWithMaxLen<T, CAPACITY> {
     type Error = RpcError;
 
     fn try_from(value: Vec<T>) -> Result<Self, Self::Error> {
--- a/libs/types/src/lib.rs
+++ b/libs/types/src/lib.rs
@@ -21,7 +21,7 @@ pub use rpc_client::{
     RegexSubstitution, RoundingError, RpcAccess, RpcAuth, RpcConfig, RpcEndpoint, RpcError,
     RpcResult, RpcSource, RpcSources, SolanaCluster, SupportedRpcProvider, SupportedRpcProviderId,
 };
-use serde::Serialize;
+use serde::{Serialize, Serializer};
 pub use solana::{
     account::{AccountData, AccountEncoding, AccountInfo, ParsedAccount},
     request::{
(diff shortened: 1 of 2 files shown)


libs/types/src/lib.rs
```diff
@@ -44,10 +44,10 @@ pub use solana::{
 
 /// A vector with a maximum capacity.
 #[derive(Debug, Clone, Deserialize, Serialize, CandidType, PartialEq, Default, Into)]
```
gregorydemay, line 46: You actually don't need `Clone`, this is an artifact of how Serde is trying to [derive bounds ](https://serde.rs/container-attrs.html#into)

> Serialize this type by converting it into the specified IntoType and serializing that. This type must implement Clone and Into<IntoType>, and IntoType must implement Serialize.

since the serializer only has a reference, it's taking ownership by calling `Clone` and then trying the conversion specified in `Into`, when actually what we want is to delegate. This seems to work

```rust
/// A vector with a maximum capacity.
#[derive(Debug, Clone, Deserialize, CandidType, PartialEq, Default, Into)]
#[serde(try_from = "Vec<T>")]
pub struct VecWithMaxLen<T, const CAPACITY: usize>(Vec<T>);

impl<T, const CAPACITY: usize> TryFrom<Vec<T>> for VecWithMaxLen<T, CAPACITY> {
    type Error = RpcError;

    fn try_from(value: Vec<T>) -> Result<Self, Self::Error> {
        if value.len() > CAPACITY {
            return Err(RpcError::ValidationError(format!(
                "Expected at most {CAPACITY} items, but got {}",
                value.len()
            )));
        }
        Ok(Self(value))
    }
}

impl<T: Serialize, const CAPACITY: usize> Serialize for VecWithMaxLen<T, CAPACITY> {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        self.0.serialize(serializer)
    }
}
```

libs/types/src/tests.rs
```diff
@@ -1,48 +1,84 @@
...
+        let encoded_parsed_value = Encode!(&parsed_value)?;
+        prop_assert_eq!(
+            &encoded_value,
+            &encoded_parsed_value,
+            "Encoded value differ for {:?}",
+            value
+        );
 
-        let result = VecWithMaxLen::<String, 100>::deserialize(&serialized);
+        let decoded_text_value = Decode!(&encoded_value, VecWithMaxLen<T, CAPACITY>)?;
```
gregorydemay, line 51: nit: The variable name seems inapropriate, maybe `decoded_value`?

libs/types/src/tests.rs
```diff
@@ -1,48 +1,84 @@
...
-    #[test]
-    fn should_not_deserialize(values in prop::collection::vec(any::<String>(), 101..1000)) {
-        let serialized = json!(values);
+    fn encode_decode_roundtrip<T, const CAPACITY: usize>(value: Vec<T>) -> Result<(), TestCaseError>
+    where
+        T: Clone + CandidType + std::fmt::Debug + PartialEq + DeserializeOwned,
+    {
+        let parsed_value: VecWithMaxLen<T, CAPACITY> = TryFrom::try_from(value.clone())?;
+        let encoded_value = Encode!(&value)?;
+        let encoded_parsed_value = Encode!(&parsed_value)?;
```
gregorydemay, line 43: nit: the variable names with `parsed` seem a bit inappropriate since we are not doing parsing. maybe something like `wrapped_value`/`encoded_wrapped_value`?

libs/types/src/tests.rs
```diff
@@ -1,48 +1,84 @@
...
 
-fn arb_vec_with_capacity<const CAPACITY: usize>(
-) -> impl Strategy<Value = VecWithMaxLen<String, CAPACITY>> {
-    prop::collection::vec(any::<String>(), 0..=CAPACITY)
-        .prop_map(|values| values.try_into().unwrap())
+    fn arb_vec_with_max_size<T: Arbitrary>(max_size: usize) -> impl Strategy<Value = Vec<T>> {
+        prop::collection::vec(any::<T>(), 0..=max_size)
+    }
+
+    fn arb_vec_with_min_size<T: Arbitrary>(min_size: usize) -> impl Strategy<Value = Vec<T>> {
```
gregorydemay, line 81: nit: from the name `with_min_size` I would actually actually expect `min_size` to be in range, in which case the test `should_fail_to_decode_when_too_long` appears wrong. Maybe rename `arb_vec_with_min_size` to something like `arb_vec_with_size_greater_than`?

gregorydemay (approved):
Thanks @lpahlavi for that PR! some minor comments.

PR #113, opened 2025-05-19
feat: add client method to sign a transaction

lpahlavi:
(XC-317) Add a helper method in the `sol_rpc_client` module to signa a Solana transaction using threshold EdDSA. This method is not part of the client `struct` as it does not interact with the SOL RPC canister, but instead with the management canister. However, it takes a runtime argument to be re-usable in tests and prod code.


examples/basic_solana/src/ed25519.rs
```diff
@@ -1,25 +1,8 @@
-use crate::Ed25519KeyName;
 use ic_cdk::api::management_canister::schnorr::{
     SchnorrAlgorithm, SchnorrKeyId, SchnorrPublicKeyArgument, SchnorrPublicKeyResponse,
-    SignWithSchnorrArgument, SignWithSchnorrResponse,
 };
 use ic_ed25519::PublicKey;
-
-#[derive(Clone, Debug, PartialEq, Eq, Default)]
-pub struct DerivationPath(Vec<Vec<u8>>);
```
lpahlavi, line 9: Moved to the `sol_rpc_types` crate.

examples/basic_solana/src/lib.rs
```diff
@@ -90,26 +90,6 @@ impl From<SolanaNetwork> for SolanaCluster {
     }
 }
 
-#[derive(CandidType, Deserialize, Debug, Default, PartialEq, Eq, Clone, Copy)]
-pub enum Ed25519KeyName {
```
lpahlavi, line 94: Moved to the `sol_rpc_types` crate.

libs/client/src/threshold_signatures.rs
```diff
@@ -0,0 +1,202 @@
...
+/// )
+/// .await
+/// .unwrap();
+///
+/// assert_eq!(pubkey, pubkey!("BPebStjcgCPnWTK3FXZJ8KhqwNYLk9aubC9b4Cgqb6oE")
+/// );
+/// # Ok(())
+/// # }
+/// ```
+pub async fn get_pubkey<R: Runtime>(
```
lpahlavi, line 164: This method is not really necessary, but I thought that it's nice to provide some utility functions to make things as easy as possible for end users. As a bonus we can re-use this method in:
* the docs for `sign_transaction` above
* `basic_solana` (and use the chain code to derive child keys locally -- we could potentially even add an example in the doc here on how to derive child keys locally)
* the end-to-end tests (in an upcoming PR where the end-to-end tests will use threshold signing)

libs/client/src/threshold_signatures.rs
```diff
@@ -0,0 +1,202 @@
...
+///     signature,
+///     Ok(Signature::from_str("37HbmunhjSC1xxnVsaFX2xaS8gYnb5JYiLy9B51Ky9Up69aF7Qra6dHSLMCaiurRYq3Y8ZxSVUwC5sntziWuhZee").unwrap())
+/// );
+///
+/// // The transaction is now signed and can be submitted with the `sendTransaction` RPC method.
+/// transaction.signatures = vec![signature.unwrap()];
+/// # Ok(())
+/// # }
+/// ```
+pub async fn sign_transaction<R: Runtime>(
```
lpahlavi, line 90: This method is a building block for the `build_transaction` method which will come in an upcoming PR. For both transactions using a recent blockhash and those using a durable nonce, it is necessary to sign the transaction (using threshold signing) if we want to have a single method call, which I find would make the most sense. The alternative would be to build an unsigned transaction, and then ask the end-users to sign the transaction themselves, but I think it's nice to make it as easy as possible for them here. Especially since we anyways need this code for end-to-end tests and `basic_solana`.

libs/client/src/threshold_signatures.rs
```diff
@@ -0,0 +1,202 @@
...
+    runtime: &R,
+    transaction: &solana_transaction::Transaction,
+    key_id: Ed25519KeyId,
+    derivation_path: Option<&DerivationPath>,
+) -> RpcResult<solana_signature::Signature> {
```
lpahlavi, line 95: The reason I don't return the signed transaction here, is that this might need to be called multiple times for transactions that have multiple signatures. AFAIK we cannot currently batch threshold signature calls, so latency might be a problem here if a transaction has many signers.

libs/client/src/threshold_sig.rs
```diff
@@ -0,0 +1,202 @@
...
+use ic_cdk::api::management_canister::schnorr::{
+    SchnorrAlgorithm, SchnorrKeyId, SignWithSchnorrArgument, SignWithSchnorrResponse,
+};
+use ic_cdk::api::management_canister::schnorr::{
+    SchnorrPublicKeyArgument, SchnorrPublicKeyResponse,
+};
+use sol_rpc_types::{DerivationPath, Ed25519KeyId, RpcError, RpcResult};
+
+// Source: https://internetcomputer.org/docs/current/references/t-sigs-how-it-works/#fees-for-the-t-schnorr-production-key
+const SIGN_WITH_SCHNORR_FEE: u128 = 26_153_846_153;
```
gregorydemay, line 18: The fee is dependent of the key being used:

1. https://internetcomputer.org/docs/current/references/t-sigs-how-it-works/#fees-for-the-t-schnorr-production-key
2. https://internetcomputer.org/docs/current/references/t-sigs-how-it-works/#fees-for-the-t-schnorr-production-key

libs/client/Cargo.toml
```diff
@@ -15,14 +15,19 @@ async-trait = { workspace = true }
 candid = { workspace = true }
```
gregorydemay, line 15: nit: since this brings a lot of dependencies and this is not really a core feature, how about putting it being a feature flag on the client?

gregorydemay (commented):
Thanks @lpahlavi for this PR! some first high level comments

lpahlavi (commented):
Thanks a lot for the initial review @gregorydemay! I've incorporated all of your feedback PTAL and let me know what you think.

libs/client/src/threshold_sig.rs
```diff
@@ -0,0 +1,254 @@
...
+    transaction: &solana_transaction::Transaction,
+    key_id: Ed25519KeyId,
+    derivation_path: Option<&DerivationPath>,
+) -> RpcResult<solana_signature::Signature> {
+    let arg = SignWithSchnorrArgument {
+        message: transaction.message_data(),
+        derivation_path: derivation_path.cloned().unwrap_or_default().into(),
+        key_id: SchnorrKeyId {
+            algorithm: SchnorrAlgorithm::Ed25519,
+            name: key_id.to_string(),
```
gregorydemay, line 151: nit: It's a bit non-idiomatic to use the `Display` implementation for this, it would be better I think to have an explicit method for this, e.g.
```rust
pub fn id(&self) -> &'static str {...}
```

libs/client/src/threshold_sig.rs
```diff
@@ -0,0 +1,254 @@
...
+/// use candid::Principal;
+/// # use ic_cdk::api::management_canister::schnorr::{SchnorrPublicKeyResponse, SignWithSchnorrResponse};
+/// let runtime = IcRuntime;
+/// # let runtime = MockRuntime::same_response(SchnorrPublicKeyResponse {
+/// #     public_key: pubkey!("BPebStjcgCPnWTK3FXZJ8KhqwNYLk9aubC9b4Cgqb6oE").as_ref().to_vec(),
+/// #     chain_code: "UWbC6EgDnWEJIU4KFBqASTCYAzEiJGsR".as_bytes().to_vec(),
+/// # });
+///
+/// let key_id = threshold_sig::Ed25519KeyId::TestKey1;
+/// let derivation_path = None;
```
gregorydemay, line 98: nit: maybe the derivation path should be instantiated from a `Principal` to be more realistic?

libs/client/src/threshold_sig.rs
```diff
@@ -0,0 +1,254 @@
...
+/// # use ic_cdk::api::management_canister::schnorr::{SchnorrPublicKeyResponse, SignWithSchnorrResponse};
+/// let runtime = IcRuntime;
+/// # let runtime = MockRuntime::same_response(SchnorrPublicKeyResponse {
+/// #     public_key: pubkey!("BPebStjcgCPnWTK3FXZJ8KhqwNYLk9aubC9b4Cgqb6oE").as_ref().to_vec(),
+/// #     chain_code: "UWbC6EgDnWEJIU4KFBqASTCYAzEiJGsR".as_bytes().to_vec(),
+/// # });
+///
+/// let key_id = threshold_sig::Ed25519KeyId::TestKey1;
+/// let canister_id = Principal::from_text("un4fu-tqaaa-aaaab-qadjq-cai").unwrap();
+/// let derivation_path = threshold_sig::DerivationPath::from("some-derivation-path".as_bytes());
```
gregorydemay, line 200: nit: similarly here a derivation path derived from a principal would be more realistic, no?

libs/client/src/threshold_sig.rs
```diff
@@ -0,0 +1,254 @@
...
+/// );
+///
+/// // The transaction is now signed and can be submitted with the `sendTransaction` RPC method.
+/// transaction.signatures = vec![signature.unwrap()];
+/// # Ok(())
+/// # }
+/// ```
+pub async fn sign_transaction<R: Runtime>(
+    runtime: &R,
+    transaction: &solana_transaction::Transaction,
```
gregorydemay, line 142: I don't understand why it's taking as parameter a `solana_transaction::Transaction`, given that this type is supposed to contain the signatures. Why no taking a `&Message`?

examples/basic_solana/src/solana_wallet.rs
```diff
@@ -85,10 +90,14 @@ impl SolanaWallet {
         )
     }
 
-    pub async fn sign_with_ed25519(&self, message: &Message, signer: &SolanaAccount) -> Signature {
-        let message = message.serialize();
-        let derivation_path = &signer.derivation_path;
-        let key_name = &read_state(|s| s.ed25519_key_name());
-        Signature::from(sign_with_ed25519(message, derivation_path, key_name).await)
+    pub async fn sign_message(message: &Message, signer: &SolanaAccount) -> Signature {
```
gregorydemay, line 93: nit: what about having that method in `SolanaAccount`? E.g.

```rust
impl SolanaAccount {
  pub fn sign(message: &Message) -> Signature {...}
}
```

gregorydemay (approved):
Thanks @lpahlavi for the changes! Some minor comments but otherwise LGTM!

libs/types/src/solana/signature/mod.rs
```diff
@@ -0,0 +1,53 @@
+use candid::{CandidType, Deserialize, Principal};
```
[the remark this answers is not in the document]
gregorydemay: yes, much better, thank you!

libs/client/src/threshold_sig.rs
gregorydemay: Thanks for checking with crypto!

>Just one follow-up question: what about the local development key? Do you know how cycles management works with local dfx deployment? Should the fee be 0 there?

I would also expect 0.

lpahlavi:
@gregorydemay I realized actually that the two new methods return an `RpcResult` which is not really correct... On the one hand it's nice to re-use the same type as RPC calls, on the other hand the error values right now are just generic `RpcError::ValidationError` and it would maybe make more sense to introduce a new error type for management canister calls, or even more specifically threshold signing errors. WDYT?

PR #117, opened 2025-05-21
feat: method to extract durable nonce from account

lpahlavi:
XC-347:  Add a helper method in the `sol_rpc_client` crate to extract the durable nonce value from the response of a `getAccountInfo` RPC call.


libs/client/src/nonce/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+///     .unwrap()
+///     .unwrap();
+///
+/// let durable_nonce = extract_durable_nonce(&usdc_account);
+///
+/// assert_matches!(durable_nonce, Err(ExtractNonceError::InvalidAccountData(_)));
+/// # Ok(())
+/// # }
+/// ```
+pub fn extract_durable_nonce(account: &UiAccount) -> Result<solana_hash::Hash, ExtractNonceError> {
```
lpahlavi, line 84: There is theoretically some code in the Solana SDK to extract a nonce value from account data [here](https://github.com/dfinity/agave/blob/6b21eae8d0eea68929ed43f79469cba32efc5810/cli/src/nonce.rs#L558). However, we cannot really use it here since they use a different account type, i.e. a `ReadableAccount` which for example `solana_account::Account` implements. However the Solana client returns a `solana_account_decoder_client_types::UiAccount`. The conversion from `solana_account_decoder_client_types::UiAccount` to `solana_account::Account` is not trivial since the former has an incompatible type for the `data` field which must first be decoded. In the end it is much easier to just write our own logic for this.

libs/client/src/nonce/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+        .map_err(|e| ExtractNonceError::InvalidAccountData(e.to_string()))?;
+    match versions.state() {
+        State::Uninitialized => Err(ExtractNonceError::Uninitialized),
+        State::Initialized(data) => Ok(data.blockhash()),
+    }
+}
+
+/// Errors that might happen when calling the [`extract_durable_nonce`] method.
+#[derive(Clone, Debug, PartialEq, Error)]
+pub enum ExtractNonceError {
```
lpahlavi, line 99: We could theoretically use the `solana_rpc_client_nonce_utils::Error` type here, but IMO it doesn't make sense to use re-use that type with potentially different meaning than originally intended.

libs/client/Cargo.toml
```diff
@@ -12,21 +12,24 @@ include = ["src", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md"]
 
 [dependencies]
 async-trait = { workspace = true }
+bincode = { workspace = true }
```
lpahlavi, line 15: I could also add a feature flag for the code in this PR. WDYT?

libs/client/src/nonce/mod.rs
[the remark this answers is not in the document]
gregorydemay: I agree, I would avoid `RpcResult`.

libs/client/Cargo.toml
```diff
@@ -12,21 +12,24 @@ include = ["src", "Cargo.toml", "CHANGELOG.md", "LICENSE", "README.md"]
...
 candid = { workspace = true }
 derive_more = { workspace = true }
 ic-cdk = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 solana-account = { workspace = true }
-solana-account-decoder-client-types = { workspace = true }
-solana-clock = { workspace = true }
+solana-account-decoder-client-types = { workspace = true, features = ["zstd"] }
```
gregorydemay, line 22: understanding question: why do we need `zstd`?

gregorydemay (commented):
Some high level understanding questions

libs/client/src/nonce/mod.rs
lpahlavi: I actually think `solana_account::Account` would be nice BUT there is a caveat...

For some absolutely unknown reason, `solana_account::Account` does not include the `space` field:

```rust
pub struct Account {
    /// lamports in the account
    pub lamports: u64,
    /// data held in this account
    #[cfg_attr(feature = "serde", serde(with = "serde_bytes"))]
    pub data: Vec<u8>,
    /// the program that owns this account. If executable, the program that loads this account.
    pub owner: Pubkey,
    /// this account's data contains a loaded program (and is now read-only)
    pub executable: bool,
    /// the epoch at which this account will next owe rent
    pub rent_epoch: Epoch,
}
```

Whereas the `UiAccount` account type does...

```rust
pub struct UiAccount {
    pub lamports: u64,
    pub data: UiAccountData,
    pub owner: String,
    pub executable: bool,
    pub rent_epoch: u64,
    pub space: Option<u64>,
}
```

Which means theoretically, we're leaving out some information by returning the `solana_account::Account` type. However, it is much easier to work with as you saw e.g. when extracting a nonce. Is this a trade-off we're willing to make?

libs/client/src/fixtures/mod.rs
```diff
@@ -116,3 +116,15 @@ pub fn usdc_account() -> AccountInfo {
         space: 82,
     }
 }
+
+/// Nonce account [`8DedqKHx9ogFajbHtRnTM3pPr3MRyVKDtepEpUiaDXX`](https://solscan.io/token/8DedqKHx9ogFajbHtRnTM3pPr3MRyVKDtepEpUiaDXX?cluster=devnet) on Solana Devnet.
```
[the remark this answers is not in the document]
lpahlavi: True! Done.

libs/client/src/nonce/mod.rs
gregorydemay: > For some absolutely unknown reason, solana_account::Account does not include the space field:

good catch!

However, I think I would still go for (`solana_account::Account`) for the following reasons:

1. AFAICT the [Solana RPC client](https://docs.rs/solana-client/latest/solana_client/rpc_client/struct.RpcClient.html#method.get_account) only returns `solana_account::Account` and not `UiAccount`
2. The `space ` field can actually be [computed](https://github.com/anza-xyz/agave/blob/e3f99e6391bdd42a2420af0bbbfcf3961d084e7b/account-decoder/src/lib.rs#L55) as `let space = account.data().len();` so that it's actually redundant.

WDYT?

lpahlavi: Yes, those are some good points. Generally I agree, and it will make our life easier to line up with the Solana RPC client. There is a small catch though.

> The space  field can actually be [computed](https://github.com/anza-xyz/agave/blob/e3f99e6391bdd42a2420af0bbbfcf3961d084e7b/account-decoder/src/lib.rs#L55) as let space = account.data().len(); so that it's actually redundant.

Yes, but, only as long as the `data_slice` parameter is not set in the request, otherwise only a subset of the account data is returned. I guess in the event the account data is very large and one does not want to fetch the whole account data at once, this could potentially be problematic. I could imagine a scenario where a user fetches the account data in batches and needs the space field to be able to tell how many batches are required... Or potentially polling if the account data size has changed. These scenarios make especially sense on the IC since the response size is limited. However I don't know how often this is actually a problem. WDYT?

lpahlavi: As discussed offline, in light of the above observation, I have kept the return type of our client to be a `solana_account_decoder_client_types::UiAccount`. I resorted to using `solana_account::Account::decode` to convert from a `solana_account_decoder_client_types::UiAccount` to an `solana_account::Account` and thus still be able to use the `solana_rpc_client_nonce_utils` crate.

libs/client/src/nonce/mod.rs
lpahlavi: @gregorydemay Since I use the `solana_rpc_client_nonce_utils` crate now, I did make use of their error type but wrapped it in our own error type. PTAL.

libs/client/Cargo.toml
lpahlavi: Yes! Done.

libs/client/src/nonce/mod.rs
lpahlavi: As discussed offline, after further analysis it seems the `solana_rpc_client_nonce_utils` crate has some dependencies that make it unusable in WASM. For this reason, I have reverted to an earlier state for this PR.

gregorydemay (approved):

PR #118, opened 2025-05-21
chore: add Chainstack RPC provider

lpahlavi:
(XC-354) Add Chainstack as a supported RPC provider for both Mainnet ande Devnet. It is currently a non-default provider for both Solana clusters.

### DIFF
--- a/canister/src/providers/mod.rs
+++ b/canister/src/providers/mod.rs
@@ -48,6 +48,24 @@ thread_local! {
                 public_url: Some("https://rpc.ankr.com/solana_devnet/".to_string()),
             }
         },
+        SupportedRpcProviderId::ChainstackMainnet => SupportedRpcProvider {
+            cluster: SolanaCluster::Mainnet,
+            access: RpcAccess::Authenticated {
+                auth: RpcAuth::UrlParameter {
+                    url_pattern: "https://solana-mainnet.core.chainstack.com/{API_KEY}".to_string(),
+                },
+                public_url: None,
+            }
+        },
+        SupportedRpcProviderId::ChainstackDevnet => SupportedRpcProvider {
+            cluster: SolanaCluster::Devnet,
+            access: RpcAccess::Authenticated {
+                auth: RpcAuth::UrlParameter {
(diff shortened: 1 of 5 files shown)


lpahlavi:
I tested the new provider manually with `dfx` with the `getSlot` method and with an API key from a personal test account:

```shell
GET_SLOT_PARAMS="(
  variant { Custom = vec { variant { Supported = variant { ChainstackMainnet } } } },
  null,
  opt record { minContextSlot = null; commitment = opt variant { finalized } },
)"
CYCLES=$(dfx canister call sol_rpc getSlotCyclesCost "$GET_SLOT_PARAMS" $FLAGS --output json | jq '.Ok' --raw-output || exit 1)
GET_SLOT_OUTPUT=$(dfx canister call sol_rpc getSlot "$GET_SLOT_PARAMS" $FLAGS --output json --with-cycles "$CYCLES" || exit 1)
SLOT=$(jq --raw-output '.Consistent.Ok' <<< "$GET_SLOT_OUTPUT")
```

And got a valid output:
```shell
(variant { Consistent = variant { Ok = 341_511_580 : nat64 } })
```

canister/src/providers/mod.rs
```diff
@@ -48,6 +48,24 @@ thread_local! {
                 public_url: Some("https://rpc.ankr.com/solana_devnet/".to_string()),
             }
         },
+        SupportedRpcProviderId::ChainstackMainnet => SupportedRpcProvider {
+            cluster: SolanaCluster::Mainnet,
+            access: RpcAccess::Authenticated {
+                auth: RpcAuth::UrlParameter {
```
gregorydemay, line 54: No header authentication?

gregorydemay (approved):
Thanks for this PR!

canister/src/providers/mod.rs
lpahlavi: @gregorydemay Seems like [not](https://docs.chainstack.com/docs/authentication-methods-for-different-scenarios#header-authentication-bearer-token). I tried it out with the existing API key but no luck...

PR #120, opened 2025-05-21
chore: add Tatum RPC provider

lpahlavi:
(XC-354) Add Tatum as a supported RPC provider for both Mainnet ande Devnet. It is currently a non-default provider for both Solana clusters.


lpahlavi:
Unfortunately, due to Tatum's low rate limit (3RPS) for the free-tier, I was not able to get a successful response for any call, however it seems that the new `HttpHeader` authentication does work with them as I get different error messages once I set an API key from a free account.

PR #124, opened 2025-05-23
test: durable nonce end-to-end test

lpahlavi:
(XC-339) Add an end-to-end test for `sendTransaction` using a durable nonce instead of a recent blockhash to construct the Solana transaction.

### DIFF
--- a/end_to_end_tests/src/lib.rs
+++ b/end_to_end_tests/src/lib.rs
@@ -143,38 +143,6 @@ impl Setup {
             .expect_consistent()
             .unwrap_or_else(|_| panic!("Failed to fetch account balance for account {pubkey}"))
     }
-
-    pub async fn get_transaction_fee(&self, transaction_id: &Signature, slot: u64) -> u64 {
-        let block = self
-            .client()
-            .get_block(slot)
-            .send()
-            .await
-            .expect_consistent()
-            .expect("Call to `getBlock` failed`")
-            .unwrap_or_else(|| panic!("No block found for slot {:?}", slot));
-        let transaction_fees = block
-            .transactions
-            .expect("Block has no transactions")
-            .into_iter()
-            .find(|transaction| {
-                transaction
-                    .transaction
-                    .decode()
-                    .expect("Failed to decode transaction")
-                    .signatures
-                    .first()
-                    == Some(transaction_id)
-            })
-            .map(|transaction| transaction.meta.expect("Transaction has no metadata").fee)
-            .unwrap_or_else(|| {
-                panic!(
-                    "Unable to get transaction fee for transaction '{:?}' from block '{:?}'",
-                    transaction_id, block.blockhash
-                )
-            });
-        transaction_fees
-    }
 }
 
 impl Default for Setup {
--- a/end_to_end_tests/src/lib.rs
+++ b/end_to_end_tests/src/lib.rs
@@ -16,6 +16,7 @@ use sol_rpc_types::{
 use solana_commitment_config::CommitmentConfig;
 use solana_pubkey::Pubkey;
 use solana_signature::Signature;
+use solana_transaction_status_client_types::TransactionStatus;
 use std::{env, time::Duration};
 
 const DEFAULT_IC_GATEWAY: &str = "https://icp0.io";
@@ -71,7 +72,7 @@ impl Setup {
             .build()
     }
 
-    pub async fn confirm_transaction(&self, transaction_id: &Signature) {
+    pub async fn confirm_transaction(&self, transaction_id: &Signature) -> TransactionStatus {
         let mut num_trials = 0;
         loop {
             num_trials += 1;
@@ -85,12 +86,12 @@ impl Setup {
                 .send()
                 .await;
             if let MultiRpcResult::Consistent(Ok(statuses)) = statuses {
-                if let Some(Some(status)) = statuses.first() {
+                if let Some(Some(status)) = statuses.into_iter().next() {
                     if let Some(err) = &status.err {
                         panic!("Transaction failed with error {:?}", err);
                     }
                     if status.satisfies_commitment(CommitmentConfig::confirmed()) {
-                        return;
+                        return status;
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


gregorydemay (commented):
Thank you @lpahlavi for that PR! Very nice to be able to test both cases. Some minor comments and refactoring suggestions from my side.

lpahlavi (commented):
Thanks a lot for the review @gregorydemay! I should have addressed/answered all  of your comments, please let me know what you think.

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -1,130 +1,86 @@
...
-        Hash::from_str(&block.blockhash).expect("Failed to parse blockhash")
-    };
-
-    let modify_instructions = |instructions: &mut Vec<Instruction>| {
-        // Set a CU limit for instructions to: perform a SOL transfer, set the CU price, and set
-        // the CU limit (150 CU x 3 = 450 CU)
-        let set_cu_limit_ix = ComputeBudgetInstruction::set_compute_unit_limit(450);
-        instructions.insert(0, set_cu_limit_ix);
+    let create_message = CreateMessageWithRecentBlockhash {
+        setup: Setup::new(),
```
gregorydemay, line 43: nit: I think here it's a bit unexpected that this takes ownership of a full setup, typically a setup is used once for the whole duration of the test, while we may want to create multiple messages within the same test. I think it should either take a reference to a `Setup` (you can wrap it in an `Arc<_>` to avoid having to deal with lifetime parameters) or a client.

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -137,37 +93,9 @@ async fn send_transaction_test<F, S>(
     let recipient_balance_before = setup.fund_account(&recipient_pubkey, 1_000_000_000).await;
```
gregorydemay, line 93: here for a lack of a better place: 

```
let setup = Setup::new();
```

we should not need a new setup. We should reuse the one we have in the current test.


nit 2:  it's also a bit weird that the method `send_transaction_test` also does `fund_account`. I think I would put this part
```
let sender_balance_before = setup.fund_account(&sender_pubkey, 1_000_000_000).await;
    println!("Sender balance before sending transaction: {sender_balance_before:?} lamports");
    let recipient_balance_before = setup.fund_account(&recipient_pubkey, 1_000_000_000).await;
println!("Recipient balance before sending transaction: {recipient_balance_before:?} lamports");
```
in some a separate helper method, like `fund_accounts` and then in `send_transaction_test` just call `get_balance` before doing the transaction.
WDYT?

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -1,38 +1,141 @@
...
+// `NONCE_ACCOUNT_B` is an initialized nonce account with nonce authority `PUBKEY_B`
+const NONCE_ACCOUNT_B: &str = "876vg5npuF9LCfc2MVWZtewBUEfcgzdbahCK7gXn5MLh";
+
+#[tokio::test]
 async fn should_send_transaction_with_recent_blockhash() {
-    let setup = Setup::new();
-    let client = setup.client();
+    let sender_pubkey = Pubkey::from_str(ACCOUNT_A).unwrap();
+    let sender_derivation_path = DerivationPath::from(DERIVATION_PATH_A);
+    verify_pubkey(&sender_derivation_path, &sender_pubkey).await;
```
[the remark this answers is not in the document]
gregorydemay: > The point of verifying the sender pubkey is that we also need to know the derivation path is correct so that threshold signing works.

Not only, you also want that the derived public key stays the same, which is what verifying the receiver public key would check. But as you said 

> To be fair though, the receiver pubkey in this test is the sender pubkey in the other test so we also verify the pubkey. 

which means we verify both, so that's good!

gregorydemay (commented):
Thanks @lpahlavi for the improvements! A minor point is that we should use at most 1 `Setup` per test, but otherwise LGTM!

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -1,130 +1,86 @@
+use async_trait::async_trait;
```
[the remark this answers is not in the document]
lpahlavi: Good idea. Done.

end_to_end_tests/tests/end_to_end.rs
```diff
@@ -52,21 +70,26 @@ async fn should_send_transaction_with_recent_blockhash() {
     .await;
 }
 
-#[tokio::test]
+#[tokio::test(flavor = "multi_thread")]
```
gregorydemay, line 73: understanding question: why is `multi_thread` needed when the e2e tests are run sequentially (`--test-threads 1`)?

end_to_end_tests/src/lib.rs
```diff
@@ -122,23 +130,57 @@ impl Setup {
...
         if balance < amount {
-            self.airdrop(account, amount).await
-        } else {
-            balance
+            self.solana_client
+                .request_airdrop(account, amount)
+                .expect("Failed to request airdrop");
+            self.solana_client.wait_for_balance_with_commitment(
+                account,
+                Some(amount),
```
gregorydemay, line 145: Shouldn't that be `balance + amount`? Note that `wait_for_balance_with_commitment` only returns `Ok(_)` when the [*exact* balance](https://github.com/dfinity/agave/blob/9d99617ff0b7fbcbf335bfd7425654d2a5b14c1d/rpc-client/src/nonblocking/rpc_client.rs#L4421) is encountered

gregorydemay (approved):
Thanks @lpahlavi !

end_to_end_tests/tests/end_to_end.rs
gregorydemay: Thanks for the changes, LGTM!

end_to_end_tests/tests/end_to_end.rs
lpahlavi: Ah yes, good question. This is a bit stupid but the Solana RPC client requires this, otherwise it complains at runtime.

PR #133, opened 2025-05-27
chore: add client builder method to set consensus strategy

lpahlavi:
Add a new `with_consensus_strategy` method to `ClientBuilder`.


gregorydemay (approved):
Thanks for these improvements (there is a superfluous import, hence lint is not happy)

PR #134, opened 2025-05-28
docs: add request parameter limitations

lpahlavi:
Some RPC methods have only limited support for the request parameters (i.e. not all fields or values of the request parameters are supported). Document this in the README.


README.md
```diff
@@ -125,21 +125,22 @@ The SOL RPC canister reaches the Solana JSON-RPC providers using [HTTPS outcalls
...
+| [`getSlot`](https://solana.com/de/docs/rpc/http/getslot)                                        | :hammer_and_wrench:  | <ul><li>The result is rounded down (configurable by caller)</li></ul>                                                                                                                                       |
+| [`getTokenAccountBalance`](https://solana.com/de/docs/rpc/http/gettokenaccountbalance)          | :scissors:           | <ul><li>The field `context` is removed from the response</li></ul>                                                                                                                                          |
+| [`getTransaction`](https://solana.com/de/docs/rpc/http/gettransaction)                          | :small_blue_diamond: | <ul><li>The `rewards` and `encoding` request parameters are not supported.</li></ul><ul><li>Only the `signatures` and `none` values for the `transactionDetails` request parameter are supported.</li></ul> |
```
gregorydemay, line 142: This does not seem correct: unless I'm missing something, there is no `rewards` request parameter in `https://solana.com/de/docs/rpc/http/gettransaction` and for `encoding` we do [support](https://github.com/dfinity/sol-rpc-canister/blob/0fdddfc30c7c8c41a37a26e3d9ae295efa9d5749/libs/types/src/solana/request/mod.rs#L411) `Base64` or `Base58`. How hard would it be to do [XC-343](https://dfinity.atlassian.net/browse/XC-343)?

> `transactionDetails`

that's supposed to be for `getBlock`, no? Similar question, how hard would it be to do [XC-342](https://dfinity.atlassian.net/browse/XC-342)?

lpahlavi: Absolutely! I think I switched the limitations for `getBlock` and `getTransaction`.

> How hard would it be to do [XC-343](https://dfinity.atlassian.net/browse/XC-343)?

> Similar question, how hard would it be to do [XC-342](https://dfinity.atlassian.net/browse/XC-342)?

Not incredibly hard I believe, although somewhat tedious since we would need to introduce corresponding types for the new return values. One potential issue though would be that the JSON format seems to have limited support from the Solana SDK. See for example [here](https://github.com/anza-xyz/agave/blob/f64b5768975fb980cb5f185ebd0a6726e59984f7/transaction-status-client-types/src/lib.rs#L169).

lpahlavi: I think adding support for `rewards` for `getBlock` should be quote easy though. We should even already have all the required types defined.

README.md
```diff
@@ -125,21 +125,22 @@ The SOL RPC canister reaches the Solana JSON-RPC providers using [HTTPS outcalls
 The limitations described above imply that it is sometimes necessary to adapt a raw response from a Solana endpoint to increase the likelihood of nodes reaching consensus when querying that endpoint using [HTTPS outcalls](https://internetcomputer.org/https-outcalls).
 The table below summarizes the supported endpoints and the necessary changes (if any) made to the response indicated as follows:
 * :white_check_mark: no changes are made to the raw response (excepted for JSON canonicalization).
+* :small_blue_diamond: one or several fields are not supported in the request parameters.
```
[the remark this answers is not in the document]
lpahlavi: Ack, I generalized the scissors to mean either removing something from the response or not supporting a request parameter.

PR #135, opened 2025-05-30
feat: add support for `rewards` parameter for `getBlock`

lpahlavi:
(XC-342) Add support for `rewards` request parameter for `getBlock` RPC method.

### DIFF
--- a/canister/sol_rpc_canister.did
+++ b/canister/sol_rpc_canister.did
@@ -411,7 +411,9 @@ type ConfirmedBlock = record {
     signatures: opt vec Signature;
     // Array of rewards distributed in this block.
     rewards: opt vec Reward;
-    // Number of partitions over which epoch rewards are distributed in this block.
+    // The epoch rewards are distributed over `1 + numRewardPartition` blocks. See the
+    // [Partitioned Inflationary Rewards Distribution](https://docs.anza.xyz/proposals/partitioned-inflationary-rewards-distribution/)
+    // feature for more details.
     numRewardPartition: opt nat64;
 };
 
--- a/libs/types/src/solana/mod.rs
+++ b/libs/types/src/solana/mod.rs
@@ -50,7 +50,9 @@ pub struct ConfirmedBlock {
     pub signatures: Option<Vec<Signature>>,
     /// Array of rewards distributed in this block.
     pub rewards: Option<Vec<Reward>>,
-    /// Number of partitions over which epoch rewards are distributed in this block.
+    /// The epoch rewards are distributed over `1 + num_reward_partitions` blocks. See the
+    /// [Partitioned Inflationary Rewards Distribution](https://docs.anza.xyz/proposals/partitioned-inflationary-rewards-distribution/)
+    /// feature for more details.
     #[serde(rename = "numRewardPartition")]
     pub num_reward_partitions: Option<u64>,
 }


libs/client/src/request/mod.rs
```diff
@@ -277,7 +277,25 @@ impl<R> GetBlockRequestBuilder<R> {
...
+    }
+
+    /// Change the `rewards` parameter for a `getBlock` request to `false`.
+    pub fn without_rewards(mut self) -> Self {
+        self.request.params.rewards = Some(false);
+        self.update_cycles()
+    }
+
+    /// Update the cycles estimate for this request
+    pub fn update_cycles(self) -> Self {
```
lpahlavi, line 290: I created a new ticket [here](https://dfinity.atlassian.net/browse/XC-385) to fix this issue with other requests as well as I feel it doesn't belong to this PR. The problem is when we set a request parameter with e.g. `with_transaction_details`, this doesn't update the cycles cost estimate. IMO we should instead only compute the cycles estimate once when sending the request (if it wasn't set manually with `with_cycles`), but we can discuss this on the new PR.

canister/scripts/examples.sh
```diff
@@ -57,6 +57,7 @@ GET_BLOCK_PARAMS="(
     commitment = opt variant { finalized };
     transactionDetails = opt variant { signatures };
     maxSupportedTransactionVersion = opt (0 : nat8);
+    rewards = null;
```
gregorydemay, line 60: nit: maybe `opt false` would be more idiomatic?

canister/sol_rpc_canister.did
```diff
@@ -403,6 +409,10 @@ type ConfirmedBlock = record {
     // The signatures of the transactions in this block. Included if the request parameter `transactionDetails` is not
     // `none`.
     signatures: opt vec Signature;
+    // Array of rewards distributed in this block.
+    rewards: opt vec Reward;
+    // Number of partitions over which epoch rewards are distributed in this block.
+    numRewardPartition: opt nat64;
```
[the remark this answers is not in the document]
gregorydemay: The API doc of [`getBlock`](https://solana.com/docs/rpc/http/getblock) is really lacking for the response, no `signatures` field, no `rewards` field, no `num_reward_partitions` field. Maybe it's worth to open an issue upstream, WDYT?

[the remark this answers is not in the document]
gregorydemay: Btw I found something, digging through the Agave Github repo:

1. The feature that introduced this field is [Partitioned Inflationary Rewards Distribution](https://docs.anza.xyz/proposals/partitioned-inflationary-rewards-distribution/)
2. Description of some behaviour https://github.com/anza-xyz/agave/issues/3619

I think the relevant part in particular is this
> With the response, we see `numRewardPartitions: 129`. [...]. We add 1 since the first block will distributes the vote account rewards, and **the 129 next blocks distribute the stake account rewards for a total of 130 blocks**

meaning that the current description 
> Number of partitions over which epoch rewards are distributed in this block.

does seem incorrect (it will span `1 + numRewardPartitions` blocks). I think we could also add a link to the feature as part of the docs. WDYT?

canister/src/rpc_client/json/mod.rs
```diff
@@ -129,16 +129,20 @@ impl GetBlockParams {
             .as_ref()
             .and_then(|config| config.transaction_details.as_ref())
     }
+
+    pub fn get_rewards(&self) -> Option<bool> {
```
gregorydemay, line 133: nit: since it returns a boolean, how about renaming `get_rewards` to something like `has_rewards` or `include_rewards`?

integration_tests/tests/solana_test_validator.rs
```diff
@@ -193,13 +191,19 @@ async fn should_get_account_info() {
...
 async fn should_not_get_account_info() {
     let setup = Setup::new().await;
-    let pubkey = Pubkey::from_str("EPjFWdd5AufqSSqeM2qN1xzybapC8G4wEGGkZwyTDt1v").unwrap();
 
     let (sol_res, ic_res) = setup
         .compare_client(
-            |sol| solana_rpc_client_get_account(&pubkey, sol, None),
+            |sol| {
+                sol.get_account_with_commitment(
+                    &system_program::id(),
```
gregorydemay, line 199: I don't understand this change, how is this test then different from `should_get_account_info`? IIUC The goal here is to query the address of an account that doesn't exist (maybe we should assert that both results are `None`, while in the other test we should assert that they are equal and `Some(_)`)

integration_tests/tests/solana_test_validator.rs
```diff
@@ -530,7 +529,7 @@ async fn should_get_signatures_for_address() {
         .compare_client(
             |sol| {
                 sol.get_signatures_for_address_with_config(
-                    &Pubkey::default(), // address of the Solana system program
+                    &system_program::id(), // address of the Solana system program
```
gregorydemay, line 532: nit: now it's self explanatory and the comment is IMO redundant.
```suggestion
                    &system_program::id(),
```

libs/client/src/lib.rs
```diff
@@ -390,13 +389,72 @@ impl<R> SolRpcClient<R> {
     }
 
     /// Call `getBlock` on the SOL RPC canister.
+    ///
+    /// # Examples
```
gregorydemay, line 393: Thanks for adding the missing Rust docs!

libs/client/src/request/mod.rs
gregorydemay: Thanks for creating the ticket!
>IMO we should instead only compute the cycles estimate once when sending the request (if it wasn't set manually with with_cycles),

yes that makes sense.

nit: I would a priori keep this new method private.

libs/types/src/solana/mod.rs
```diff
@@ -48,6 +48,11 @@ pub struct ConfirmedBlock {
     /// Signatures of the transactions in the block. Included in the response whenever
     /// `transactionDetails` is not `none`.
     pub signatures: Option<Vec<Signature>>,
+    /// Array of rewards distributed in this block.
+    pub rewards: Option<Vec<Reward>>,
+    /// Number of partitions over which epoch rewards are distributed in this block.
```
gregorydemay, line 53: nit: also update the Rust doc here according to first comment on Candid interface.

gregorydemay (commented):
Thanks @lpahlavi for this PR! Only some minor comments on my side.

integration_tests/tests/solana_test_validator.rs
lpahlavi: Good catch, I blindly changed `Pubkey::default()` to `system_program::id()` but actually the test was wrong before. It should now be fixed. I also added an `assert_matches!(..., Some(_))` in the other test.

libs/client/src/request/mod.rs
lpahlavi: Good point, unfortunately it is called in the client when initializing a request [here](https://github.com/dfinity/sol-rpc-canister/blob/2f3df1c7b33dd68b22509524a177ee1fdff8039a/libs/client/src/lib.rs#L457) so it cannot be private... Another reason why I think it would be better to refactor where the cycles calculation takes place (in another PR).

integration_tests/tests/solana_test_validator.rs
gregorydemay: Thanks for the fix!

gregorydemay (approved):

PR #136, opened 2025-06-02
chore: add helper methods for request builders

lpahlavi:
Add explicit types for all RPC request builders, and methods to set optional request parameters (e.g. `with_min_context_slot()`, `.with_commitment()`).

### DIFF
--- a/libs/client/src/request/mod.rs
+++ b/libs/client/src/request/mod.rs
@@ -442,17 +442,13 @@ pub type GetSlotRequestBuilder<R> = RequestBuilder<
 impl<R> GetSlotRequestBuilder<R> {
     /// Change the `commitment` parameter for a `getSlot` request.
     pub fn with_commitment(mut self, commitment_level: CommitmentLevel) -> Self {
-        let mut config = self.request.params.unwrap_or_default();
-        config.commitment = Some(commitment_level);
-        self.request.params = Some(config);
+        self.request.params.get_or_insert_default().commitment = Some(commitment_level);
         self
     }
 
     /// Change the `minContextSlot` parameter for a `getSlot` request.
     pub fn with_min_context_slot(mut self, slot: Slot) -> Self {
-        let mut config = self.request.params.unwrap_or_default();
-        config.min_context_slot = Some(slot);
-        self.request.params = Some(config);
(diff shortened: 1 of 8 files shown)


libs/types/src/solana/request/mod.rs
```diff
@@ -7,7 +7,7 @@ use candid::{CandidType, Deserialize};
 use serde::Serialize;
 
 /// The parameters for a Solana [`getAccountInfo`](https://solana.com/docs/rpc/http/getaccountinfo) RPC method call.
-#[derive(Debug, Clone, Deserialize, Serialize, CandidType)]
+#[derive(Debug, Clone, Deserialize, Serialize, CandidType, PartialEq)]
```
gregorydemay, line 10: nit: we usually try to follow the following conventions for deriving traits

1. Rust standard trait (often by alphabetical order)
2. Followed by non-standard traits (often by alphabetical order)



```suggestion
#[derive(Clone, Debug, PartialEq, CandidType, Deserialize, Serialize)]
```

libs/client/src/request/mod.rs
```diff
@@ -351,6 +431,32 @@ impl SolRpcRequest for GetSlotRequest {
...
+    MultiRpcResult<Slot>,
+    MultiRpcResult<Slot>,
+>;
+
+impl<R> GetSlotRequestBuilder<R> {
+    /// Change the `commitment` parameter for a `getSlot` request.
+    pub fn with_commitment(mut self, commitment_level: CommitmentLevel) -> Self {
+        let mut config = self.request.params.unwrap_or_default();
+        config.commitment = Some(commitment_level);
+        self.request.params = Some(config);
```
gregorydemay, line 447: nit: here and below, you can simplify this with [`get_or_insert_default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_default):
```suggestion
        let mut params = self.request.params.get_or_insert_default();
        config.commitment = Some(commitment_level);
```

gregorydemay (approved):
Thanks a lot @lpahlavi for consolidating the client API!

libs/types/src/solana/request/mod.rs
lpahlavi: Ah yes, thanks for pointing that out! Done.

PR #140, opened 2025-06-06
refactor: change `Runtime` trait to use `&mut self` references

lpahlavi:
(XC-317) Change the `Runtime` `query_call` and `update_call` methods to take a `&mut self` argument instead of a `&self`.

### DIFF
--- a/libs/client/src/lib.rs
+++ b/libs/client/src/lib.rs
@@ -125,6 +125,8 @@ pub mod fixtures;
 pub mod nonce;
 mod request;
 
+#[cfg(feature = "ed25519")]
+use crate::ed25519::{get_pubkey, sign_message, DerivationPath, Ed25519KeyId};
 use crate::request::{
     GetAccountInfoRequest, GetAccountInfoRequestBuilder, GetBalanceRequest,
     GetBalanceRequestBuilder, GetBlockRequest, GetBlockRequestBuilder,
@@ -137,6 +139,7 @@ use crate::request::{
 };
 use async_trait::async_trait;
 use candid::{utils::ArgumentEncoder, CandidType, Principal};
+use futures::lock::Mutex;
 use ic_cdk::api::call::RejectionCode;
 pub use request::{Request, RequestBuilder, SolRpcEndpoint, SolRpcRequest};
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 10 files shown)


libs/client/src/lib.rs
```diff
@@ -222,9 +230,9 @@ impl SolRpcClient<IcRuntime> {
 }
 
 /// Client to interact with the SOL RPC canister.
-#[derive(Clone, Eq, PartialEq, Debug)]
+#[derive(Debug)]
 pub struct ClientConfig<R> {
-    runtime: R,
```
lpahlavi, line 227: @gregorydemay The core of the difficulty is here... Since the `SolRpcClient<R>` has an `Arc<ClientConfig<R>>`, I needed to change this to a mutex (it doesn't really make sense IMO to store the whole config in a mutex since otherwise we need to call `lock` everytime we e.g. want the `sol_rpc_canister`). Not sure about the performance cost, maybe you have a better idea than me? Since this is just needed for testing (at least thus far), I don't know if it is acceptable. WDYT?

gregorydemay: good question, I would like to avoid `Mutex` as much as possible, there is a runtime cost associated to it that doesn't seem to be warranted here (plus it potentially locks any other call in case of bugs).

I didn't try it out but maybe you could use `std::mem::replace` like described [here](https://docs.rs/tower/latest/tower/trait.Service.html)

```rust
let clone = self.inner.clone();
// take the service that was ready
let mut inner = std::mem::replace(&mut self.inner, clone);
inner.call(req)
```

lpahlavi: @gregorydemay Thanks for having a look! AFAIK though this doesn't work because the `ClientConfig` is inside an `Arc`. I guess what you meant is something along those lines:

```rust
pub async fn with_mut_runtime<F, S>(&mut self, f: F) -> S
where
    F: AsyncFnOnce(&mut R) -> S,
{
    let mut runtime = std::mem::replace(&mut self.config.runtime, self.config.runtime.clone());
    f(&mut runtime).await
}
```

i.e. you "take" the runtime, do something mutable on it, and probably then "put it back" somehow. This doesn't seem to work with `Arc`.

We would need a way to introduce interior mutability which, IIUC is usually done with `Arc + Mutex` for multi-threaded environments (which we are in here since the client could be used in a multi-threaded environment, especially in tests), or `Rc + Refcell` in single-threaded environments. Intuitively I think we have to "lock" the runtime one way or another if we are doing a mutable operation on it, otherwise there would be a race condition if we try to write back the mutated runtime while someone else is trying to mutate it simultaneously.

WDYT?

PR #142, opened 2025-06-10
test: integration tests for `estimate_recent_blockhash`

lpahlavi:


integration_tests/src/lib.rs
```diff
@@ -495,12 +512,16 @@ where
...
+        request_body: serde_json::Value,
+        response_body: serde_json::Value,
+    ) -> Self;
+    fn mock_json_rpc_response<const N: usize>(self, body: serde_json::Value) -> Self;
+    fn mock_json_rpc_responses<const N: usize>(self, bodies: [serde_json::Value; N]) -> Self;
     fn mock_sequential_json_rpc_responses<const N: usize>(
         self,
-        status: u16,
-        body: serde_json::Value,
+        bodies: Vec<serde_json::Value>,
```
gregorydemay, line 524: I have the impression we have a lot of methods that do similar things and the difference is not immediately clear. I have the impression we could have a single `mock` method that takes as parameter a mock and then we move these different behaviors that we need into a builder for that mock (which may contain a sequence of mocking strategies). WDYT?

PR #143, opened 2025-06-11
test: integration test for fetching metrics

lpahlavi:
(XC-296) Add integration test for fetching metrics from the SOL RPC canister HTTP endpoint.


integration_tests/src/lib.rs
```diff
@@ -223,6 +224,24 @@ impl Setup {
...
         self.sol_rpc_canister_id
     }
+
+    pub async fn check_metrics(self) -> MetricsAssert<Self> {
+        MetricsAssert::from_async_http_query(self).await
+    }
+}
+
+#[async_trait]
+impl AsyncCanisterHttpQuery<RejectResponse> for Setup {
```
gregorydemay, line 234: nit: let's try to move this to the lib `ic-metrics-assert`

PR #147, opened 2025-06-12
chore: Revisit response size estimates

lpahlavi:
(XC-338) Revisit and update response size estimates for some endpoints where the original estimates were not accurate.

### DIFF
--- a/integration_tests/tests/tests.rs
+++ b/integration_tests/tests/tests.rs
@@ -1655,7 +1655,7 @@ mod cycles_cost_tests {
                     check(
                         &setup,
                         client.get_account_info(USDC_PUBLIC_KEY),
-                        1_793_744_800,
+                        1_752_822_400,
                     )
                     .await;
                 }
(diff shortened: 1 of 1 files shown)


canister/src/rpc_client/mod.rs
lpahlavi, line 1: @gregorydemay I can't comment down there, but currently the default estimate for a generic JSON request is 1KB. I think it's a fair estimate, and end-users should set an explicit value anyways in most cases. WDYT?

canister/src/rpc_client/mod.rs
```diff
@@ -100,7 +100,7 @@ impl GetAccountInfoRequest {
         let providers = Providers::new(rpc_sources, consensus_strategy.clone())?;
         let max_response_bytes = config
             .response_size_estimate
-            .unwrap_or(1024 + HEADER_SIZE_LIMIT);
+            .unwrap_or(512 + HEADER_SIZE_LIMIT);
```
lpahlavi, line 103: @gregorydemay This one is a bit tricky as the response size can vary a lot depending on the data stored. An account with no data (e.g. just a balance such as [this one](https://explorer.solana.com/tx/8NwsC1mBp2N94jDdngGPW6Zgu6PEQbSxwwqUR7y5yTRVpLmRP39J7WrkphS1ewMrvjnSkQzwD1HwnTGcrUd4i53)) is around 256B and a nonce account (e.g. [this one](876vg5npuF9LCfc2MVWZtewBUEfcgzdbahCK7gXn5MLh)) is still below 512B. However a larger account such as the [token program account](https://explorer.solana.com/address/TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA) is closer to 100KB.

IMO it's fine to put an estimate here for a smaller account and expect users to add an explicit estimate in case they are fetching a larger account. WDYT?

PR #152, opened 2025-06-19
feat: deploy `basic_solana` on ICP Ninja

lpahlavi:
(XC-363) Support deployment of `basic_solana` to [ICP Ninja](https://icp.ninja/).

#159 is a follow-up PR that automates the generation of the `Cargo.toml` and `Cargo.lock` files for the ICP Ninja deployment.


ninegua (commented):
I get the following warning when running `cargo check --all`:
```
warning: Patch `solana-compute-budget-interface v2.2.2 (https://github.com/dfinity/solana-sdk?tag=46ca4e2-js-feature-flag#46ca4e22)` was not used in the crate graph.
```

Do you think we should also build this package as part of the CI? @lpahlavi

lpahlavi:
> I get the following warning when running `cargo check --all`:
> 
> ```
> warning: Patch `solana-compute-budget-interface v2.2.2 (https://github.com/dfinity/solana-sdk?tag=46ca4e2-js-feature-flag#46ca4e22)` was not used in the crate graph.
> ```
> 
> Do you think we should also build this package as part of the CI? @lpahlavi

Good observation @ninegua! This actually because I copy-pasted the whole `[patch.crates-io]` section from the top-level `Cargo.toml` into `examples/basic_solana/ninja/Cargo.toml` (see #159), but that specific crate (i.e. `solana-compute-budget-interface`) is not used within the `basic_solana`  example. 

IMO that's an OK tradeoff to keep the logic for generating `examples/basic_solana/ninja/Cargo.toml` simple i.e. just copy-paste the whole `[patch.crates-io]` section from the top-level `Cargo.toml` instead of picking and choosing which dependencies we use, which would have to include looking at transitive dependencies. Additionally, once the next release of `solana-sdk` libraries are published, we will finally be able to get rid of the patched libraries since they merged a PR that allows removing the dependencies on `wasm-bindgen` (see [here](https://github.com/anza-xyz/solana-sdk/pull/244)), meaning this is a temporary solution anyways.

WDYT?

ninegua (approved):

PR #156, opened 2025-06-25
chore: use build script for `basic_solana` deployment

lpahlavi:
Use a build script to deploy `basic_solana` locally and on Mainnet which checks what platform it is running to ensure proper handling of `llvm`.

### DIFF
--- a/examples/basic_solana/build.sh
+++ b/examples/basic_solana/build.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+TARGET="wasm32-unknown-unknown"
+
+# Optional manifest path
+MANIFEST_ARG=""
+if [ $# -ge 1 ]; then
+  MANIFEST_ARG="--manifest-path $1"
+fi
+
+# Build based on the platform
+if [ "$(uname)" == "Darwin" ]; then
+  LLVM_PATH=$(brew --prefix llvm)
+  AR="${LLVM_PATH}/bin/llvm-ar" CC="${LLVM_PATH}/bin/clang" cargo build --target "$TARGET" --release $MANIFEST_ARG
+else
+  cargo build --target "$TARGET" --release $MANIFEST_ARG
+fi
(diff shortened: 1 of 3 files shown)


examples/basic_solana/build.sh
```diff
@@ -0,0 +1,18 @@
...
+if [ $# -ge 1 ]; then
+  MANIFEST_ARG="--manifest-path $1"
+fi
+
+# Build based on the platform
+if [ "$(uname)" == "Darwin" ]; then
+  LLVM_PATH=$(brew --prefix llvm)
+  AR="${LLVM_PATH}/bin/llvm-ar" CC="${LLVM_PATH}/bin/clang" cargo build --target "$TARGET" --release $MANIFEST_ARG
+else
+  cargo build --target "$TARGET" --release $MANIFEST_ARG
```
gregorydemay, line 17: why not keeping the `--no-default-features` flag?

gregorydemay (approved):

examples/basic_solana/build.sh
lpahlavi: @gregorydemay Good question! 
* For SOL RPC canister code, there is actually a required default feature (namely `zstd`).  When compiling without it, it would lead to runtime errors when trying to use some functionality related to `zstd` (e.g. decoding the result of `getAccountInfo` with `base64+zstd` encoding). 
* With `basic_solana` it shouldn't make a difference, but I see no reason to add this flag.

One solution would also simply be to remove that feature in the SOL RPC canister since currently we should have no problems with `llvm` anymore.

WDYT?

examples/basic_solana/build.sh
```diff
@@ -0,0 +1,18 @@
...
+TARGET="wasm32-unknown-unknown"
+
+# Optional manifest path
+MANIFEST_ARG=""
+if [ $# -ge 1 ]; then
+  MANIFEST_ARG="--manifest-path $1"
+fi
+
+# Build based on the platform
+if [ "$(uname)" == "Darwin" ]; then
```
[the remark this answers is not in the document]
lpahlavi: Yes, I also thought it might be a little overkill, though we can keep it in mind.

PR #159, opened 2025-06-27
build: Add CI pipeline job to check Ninja `Cargo.toml`

lpahlavi:
(XC-363) Add CI pipeline jobs to:
* Check that the `Cargo.toml` file in `examples/basic_solana/ninja` is a standalone version of the one in examples/basic_solana` and includes the correct dependency versions and flags.
* Check that `Cargo.lock` and `examples/basic_solana/ninja/Cargo.lock` are up-to-date.

### DIFF
--- a/examples/basic_solana/ninja/Cargo.toml
+++ b/examples/basic_solana/ninja/Cargo.toml
@@ -10,7 +10,7 @@ path = "src/main.rs"
 
 [dependencies]
 bs58 = "0.5.1"
-candid = "0.10.13"
+candid = "0.10.14"
 # Transitive dependency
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -34,6 +34,43 @@ jobs:
       - name: 'Check Cargo.toml'
         run: cargo sort --workspace --check
 
+  check-ninja-cargo-toml:
```
gregorydemay, line 37: shouldn't we also check that we can deploy the basic_solana/ninja example locally? (To for example ensure that the init args are correct)?

PR #160, opened 2025-06-27
fix: use correct fee for t-sig with local development key

lpahlavi:
(XC-403) Until now, calls to the `sign_with_schnorr` management canister endpoint were incorrectly using the same fee for local and test keys, whereas the local key requires the same fee as the production key, see [the documentation](https://internetcomputer.org/docs/references/t-sigs-how-it-works#local-development-environment).


gregorydemay (approved):
Thanks @lpahlavi for the fix!

PR #161, opened 2025-06-27
docs: improve `basic_solana` Candid documentation

lpahlavi:
(XC-405) Improve the documentation in `basic_solana.did` for all endpoints.

### DIFF
--- a/examples/basic_solana/basic_solana.did
+++ b/examples/basic_solana/basic_solana.did
@@ -163,13 +163,19 @@ service : (InitArg) -> {
     ) -> (Txid);
 
     // Sends SPL tokens from the sender's associated token account (ATA)
-    // to a recipient Solana address's ATA.
+    // to the recipient's ATA, derived from the provided recipient Solana
+    // address and token mint.
+    //
+    // Both the sender and recipient must already have initialized ATAs for
+    // the given SPL token mint. This method does **not** create ATAs or
+    // initialize them.
     //
     // Parameters:
-    // - `owner` (optional): The principal to derive the sender’s ATA from.
+    // - `owner` (optional): The principal used to derive the sender’s ATA.
     //   If omitted, the caller's principal is used.
     // - `mint_account`: The SPL token mint address.
-    // - `to`: The recipient's Solana address (not their ATA).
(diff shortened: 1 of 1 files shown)


gregorydemay (approved):
Thanks @lpahlavi for improving the docs!

examples/basic_solana/basic_solana.did
```diff
@@ -64,75 +64,118 @@ type TokenAmount = record {
...
+    //   addresses from. If omitted, the caller's principal is used.
+    // - `to`: The recipient Solana address.
+    // - `amount`: The amount of SOL to send, in lamports.
+    send_sol_with_durable_nonce : (
+        owner: opt principal,
+        to: Address,
+        amount: Lamport
+    ) -> (Txid);
+
+    // Sends SPL tokens from the sender's associated token account (ATA)
```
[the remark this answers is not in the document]
lpahlavi: Absolutely, good catch! Done.

examples/basic_solana/basic_solana.did
```diff
@@ -64,75 +64,124 @@ type TokenAmount = record {
...
 
 service : (InitArg) -> {
-    // Returns the Solana account derived from the owner principal.
+    // Returns the Solana address derived from a principal.
     //
-    // If the owner is not set, it defaults to the caller's principal.
+    // Parameters:
+    // - `owner` (optional): The principal to derive the address from.
+    //   If omitted, the caller's principal is used.
     solana_account : (owner: opt principal) -> (Address);
```
letmejustputthishere, line 72: it's a bit confusing that the function is called `solana_account` but returns an address. should this be renamed for this and the other endpoints?

lpahlavi: @letmejustputthishere That's a good point. I merged this PR, but I can potentially open another one for renaming the endpoints. My reasoning is that pretty much everywhere in the Candid interface, accounts are identified by an address. It's a bit of a subtle difference, as the account is rather the data contained in the account, and the address is the identifier for an account. In this sense I don't see a contradiction per se, but I agree that it could be confusing... One option would be to just use address everywhere (i.e. in parameter and method names). @gregorydemay do you have an opinion?

PR #173, opened 2025-07-04
docs: add deployment instructions

lpahlavi:
(XC-405) Add section to the top-level README on how to deploy a local instance of the SOL RPC canister. Also refer to this section in the "Usage" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ Interact with the [Solana](https://solana.com/) blockchain from the [Internet Co
 * [Usage](#usage)
     * [From the command line](#from-the-command-line)
     * [From within a Rust canister](#from-within-a-rust-canister)
+* [Deployment](#deployment)
+  * [Deployment to the IC](#deployment-to-the-ic)
+  * [Local deployment](#local-deployment)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +106,40 @@ let slot = client
 
 Full examples are available in the [examples](examples) folder and additional code snippets are also available in the [`sol_rpc_client`](libs/client/README.md) crate.
 
+## Deployment
```
gregorydemay, line 109: nit: let's add `Deployment` to the Table of Contents.

I would add 2 sub-sections:

1. Deployment on Mainnet
2. Local Deployment: this one in particular should have a big fat warning along the line that the fact that it works locally does not mean that it will work on mainnet (e.g. IPv4/IPv6 and 1 replica/replicated call leading to consensus errors). Any real-world application should be tested on mainnet.

gregorydemay (commented):
Thanks @lpahlavi for the additional docs! Some minor comments.

gregorydemay (approved):
Thanks @lpahlavi for the new docs!

PR #174, opened 2025-07-04
build: release with multi-tags

gregorydemay:


release-plz.toml
```diff
@@ -8,11 +8,11 @@ changelog_update = true
 # update dependencies with `cargo update`
 dependencies_update = true
 
-# Disable git releases for all packages by default
-git_release_enable = false
+# Enable git releases for all packages by default
```
gregorydemay, line 11: @lpahlavi :I tested that it work by running the workflow. Example of created [PR](https://github.com/dfinity/sol-rpc-canister/pull/175)

lpahlavi: Understanding question: I see that the example release workflow includes releases for:
* `sol_rpc_types` v1.0.1: ✅ The latest version on [crates.io](https://crates.io/) is **not** up-to-date.
* `sol_rpc_canister` v1.0.0: ❓ Do we want to publish this on [crates.io](https://crates.io/)? I don't see any real reason not to, but we didn't publish it until now so I want to make sure this is intentional.

And no release for:
* `sol_rpc_client`: ✅ The latest version on [crates.io](https://crates.io/) is up-to-date.

lpahlavi (commented):
Thanks a lot for the PR @gregorydemay! Just an understanding question from my side.

release-plz.toml
lpahlavi: Thanks a lot for the explanations!

> Maybe it would be cleaner to revert the bump from https://github.com/dfinity/sol-rpc-canister/pull/164 so that the release is processed normally and we can check that everything works as expected. WDYT?

Good idea! [Here](https://github.com/dfinity/sol-rpc-canister/pull/178) is the PR. Let's merge it and check everything works as expected before merging this PR.

gregorydemay: @lpahlavi That's the new release #179

gregorydemay: @lpahlavi : Update with the correctly detected breaking change in `sol_rpc_types`: #181

lpahlavi (approved):
Thanks a lot for the PR @gregorydemay, LGTM!

PR #185, opened 2025-07-09
refactor: require HTTP outcall base fee

lpahlavi:
(XC-416) Require all requests to endpoints making an HTTP outcall to have at least the base cycles fee of an HTTP outcall attached when not in demo mode.

### DIFF
--- a/canister/src/memory/tests.rs
+++ b/canister/src/memory/tests.rs
@@ -224,7 +224,12 @@ mod upgrade_state_tests {
     }
 
     pub fn arb_api_key() -> impl Strategy<Value = ApiKey> {
-        VALID_API_KEY_CHARS.prop_map(|value| ApiKey::try_from(value).unwrap())
+        proptest::collection::vec(
+            prop::sample::select(VALID_API_KEY_CHARS.chars().collect::<Vec<_>>()),
+            1..=20,
+        )
+        .prop_map(String::from_iter)
+        .prop_map(|value| ApiKey::try_from(value).unwrap())
     }
 
     pub fn arb_principal() -> impl Strategy<Value = Principal> {


libs/client/src/request/mod.rs
```diff
@@ -785,15 +786,30 @@ impl<Runtime, Config, Params, CandidOutput, Output>
...
+    {
+        let rpc_method = self.request.endpoint.rpc_method();
+        self.try_send()
+            .await
+            .unwrap_or_else(|e| panic!("Client error: failed to call `{}`: {e:?}", rpc_method))
+    }
+
+    /// Constructs the [`Request`] and send it using the [`SolRpcClient`]. This method returns
+    /// either the request response or any error that occurs while sending the request.
+    pub async fn try_send(self) -> Result<Output, (RejectionCode, String)>
```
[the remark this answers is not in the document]
lpahlavi: Ok, I _think_ this shouldn't happen (see [here](https://internetcomputer.org/docs/references/ic-interface-spec#system-api-call) the description of a 0 error code), so I left the panic but updated the message slightly. WDYT?

libs/client/src/request/mod.rs
```diff
@@ -785,16 +787,46 @@ impl<Runtime, Config, Params, CandidOutput, Output>
...
+    /// either the request response or any error that occurs while sending the request.
+    pub async fn try_send(self) -> Result<Output, (RejectCode, String)>
     where
         Config: CandidType + Send,
         Params: CandidType + Send,
         CandidOutput: Into<Output> + CandidType + DeserializeOwned,
     {
         self.client
-            .execute_request::<Config, Params, CandidOutput, Output>(self.request)
+            .try_execute_request::<Config, Params, CandidOutput, Output>(self.request)
```
gregorydemay, line 816: nit: I would move the error mapping to `try_execute_request` so that basically no caller has to deal with `ic_cdk::api::call::RejectionCode`

libs/client/src/request/mod.rs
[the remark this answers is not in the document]
gregorydemay: See this [thread](https://dfinity.slack.com/archives/CGA566TPV/p1749199745556099)

libs/client/src/request/mod.rs
```diff
@@ -785,16 +787,46 @@ impl<Runtime, Config, Params, CandidOutput, Output>
...
             .await
+            .map_err(|(code, message)| {
+                let code = match code {
+                    RejectionCode::NoError => panic!("BUG: call failed without error"),
+                    RejectionCode::SysFatal => RejectCode::SysFatal,
+                    RejectionCode::SysTransient => RejectCode::SysTransient,
+                    RejectionCode::DestinationInvalid => RejectCode::DestinationInvalid,
+                    RejectionCode::CanisterReject => RejectCode::CanisterReject,
+                    RejectionCode::CanisterError => RejectCode::CanisterError,
+                    RejectionCode::Unknown => RejectCode::SysUnknown,
```
gregorydemay, line 826: This is incorrect, `RejectCode::SysUnknown` has a very specific semantic solely used for best effort messages, see [reject codes](https://internetcomputer.org/docs/references/ic-interface-spec#reject-codes):
> This code is only applicable to inter-canister calls that used ic0.call_with_best_effort_response

gregorydemay (commented):
Thanks for the improvements @lpahlavi! Some remarks regarding the error mapping.

libs/client/src/request/mod.rs
```diff
@@ -785,16 +787,46 @@ impl<Runtime, Config, Params, CandidOutput, Output>
...
         Config: CandidType + Send,
         Params: CandidType + Send,
         CandidOutput: Into<Output> + CandidType + DeserializeOwned,
     {
         self.client
-            .execute_request::<Config, Params, CandidOutput, Output>(self.request)
+            .try_execute_request::<Config, Params, CandidOutput, Output>(self.request)
             .await
+            .map_err(|(code, message)| {
+                let code = match code {
```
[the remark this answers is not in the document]
lpahlavi: I'm closing this thread since I moved the relevant refactoring to #187

canister/src/main.rs
```diff
@@ -32,6 +32,26 @@ pub fn require_api_key_principal_or_controller() -> Result<(), String> {
...
 }
 
+pub fn require_base_http_outcall_fee() -> Result<(), String> {
+    // See: https://internetcomputer.org/docs/references/cycles-cost-formulas#https-outcalls
+    fn base_fee() -> u128 {
+        let num_nodes = read_state(|state| state.get_num_subnet_nodes());
+        3_000_000_u128
+            .saturating_add(60_000_u128.saturating_mul(num_nodes as u128))
+            .saturating_mul(num_nodes as u128)
+    }
```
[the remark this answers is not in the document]
lpahlavi: Yes! Good idea! Done.

canister/src/memory/mod.rs
```diff
@@ -96,6 +96,7 @@ pub struct State {
     log_filter: LogFilter,
     mode: Mode,
     num_subnet_nodes: u32,
+    base_http_outcall_fee: u128,
```
gregorydemay, line 99: note that adding a mandatory field to a struct serialized in stable memory will fail to upgrade, as shown by the CI pipeline

> Client error: failed to call `getBalance`: (SysFatal, "An error happened during the call: 5: IC0503: Error from Canister zaylz-mqaaa-aaaar-qaqzq-cai: Canister called `ic0.trap` with message: 'Panicked at 'failed to decode memory bytes a6686170695f6b657973a96e416c6368656d794d61696e6e65747820306e6d7372394e755a4c347a57744f62544f7747467035314e776638626969346d416c6368656d794465766e65747820306e6d7372394e755a4c347a57744f62544f7747467035314e776638626969346b416e6b724d61696e6e65747840663330663736633030386361626539343130373238363130383161373862326661313865643937383630616263366432373534656563373332646236383532316a416e6b724465766e657478406633306637366330303863616265393431303732383631303831613738623266613138656439373836306162633664323735346565633733326462363835323171436861696e737461636b4d61696e6e65747820323334303131643131326336303339666531663837363038663139306662636570436861696e737461636b4465766e6574782066323562396561656635626263653031313235333639643734643434663230626a447270634465766e6574782c416d5356303941636845703874447a4243743873796a55625530524f45377752384c30464e6a57436e30326a6d48656c6975734d61696e6e6574782438623464643638612d383031322d346533342d396361352d3337356664333732323938636c48656c6975734465766e6574782438623464643638612d383031322d346533342d396361352d333735666433373232393863726170695f6b65795f7072696e636970616c7381581dce7a4082861dd14808ea4b8ac53a9f2a59e35747757afd24c2d00a7302716f766572726964655f70726f7669646572a16c6f766572726964655f75726cf66a6c6f675f66696c7465726753686f77416c6c646d6f6465664e6f726d616c706e756d5f7375626e65745f6e6f6465731822: Semantic(None, \"missing field `base_http_outcall_fee`\")', canister/src/memory/mod.rs:88:29'.\nConsider gracefully handling failures from this canister or altering the canister to handle exceptions. See documentation: https://internetcomputer.org/docs/current/references/execution-errors#trapped-explicitly")

It would be a good idea to 

1. add a test to ensure state compatibility, e.g. like [this](https://github.com/dfinity/ic/blob/642b305524735a16e030ab56d143e16240e13d53/rs/ethereum/ledger-suite-orchestrator/src/state/tests.rs#L504)
2. Make this field optional and change the getter to something like `pub fn compute_base_http_outcall_fee(&mut self) -> u128` that returns the previously computed value if there is one and otherwise set it and returns.

lpahlavi: Good catch! I've modified `State` to have an optional `base_http_outcall_fee` field instead and added some tests.

canister/src/memory/mod.rs
```diff
@@ -164,8 +167,9 @@ impl State {
         self.mode = mode
     }
 
-    pub fn get_base_http_outcall_fee(&self) -> u128 {
+    pub fn lazy_compute_base_http_outcall_fee(&self) -> u128 {
```
gregorydemay, line 170: Note that this will constantly re-evaluate `compute_base_http_outcall_fee(self.num_subnet_nodes)` if the initial value is empty. To be lazy, it's **unavoidable** that this method must be able to mutate the underlying state (i.e. parameter `&mut self` instead of `&self`).

canister/src/types/mod.rs
```diff
@@ -2,12 +2,15 @@
...
 use crate::{constants::API_KEY_REPLACE_STRING, validate::validate_api_key};
+#[cfg(test)]
+use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
 use sol_rpc_types::{RegexSubstitution, RpcEndpoint};
 use std::{fmt, fmt::Debug};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 #[derive(Clone, PartialEq, Zeroize, ZeroizeOnDrop, Deserialize, Serialize)]
+#[cfg_attr(test, derive(Arbitrary))]
```
gregorydemay, line 13: nit: I personally dislike mixing test code with production code. I'm not sure it's currently worth it to add an additional library just to derive arbitrary for `ApiKey`, when it easilly could be derived manually.

canister/src/memory/tests.rs
```diff
@@ -55,3 +67,180 @@ mod request_counter_tests {
...
+                        num_subnet_nodes,
+                    )| VersionedState::V0 {
+                        api_keys,
+                        api_key_principals,
+                        override_provider,
+                        log_filter,
+                        mode,
+                        num_subnet_nodes,
+                    }
+                ),
```
gregorydemay, line 172: nit: it would be more readable I think to extract a function out of it, e.g. `arb_state_v0` (and similarly `arb_state_v1`).

gregorydemay (commented):
Thanks for adding the test! I think there is a minor problem with the `lazy_` method but otherwise code LGTM!

PR #186, opened 2025-07-10
fix: remove unneeded `rename` annotations

lpahlavi:
Remove unneeded Candid method `rename` annotations in the SOL RPC canister.


gregorydemay (approved):
Nice! Thanks for the clean-up @lpahlavi !

PR #204, opened 2025-08-05
chore: Update README.md

letmejustputthishere:
fix path to deployment instructions


lpahlavi (approved):
Thanks for the fix @letmejustputthishere!

PR #206, opened 2025-08-08
chore: upgrade SOL RPC canister to v1.1.0

lpahlavi:
Add proposal to upgrade SOL RPC canister to v1.1.0.


mducroux (approved):
LGTM @lpahlavi!

PR #215, opened 2025-08-27
fix: release pipeline

lpahlavi:
Currently, `release-plz` runs into issues when trying to determine the next version of the `sol_rpc_canister` crate. This is because the `sol_rpc_canister` crate is a binary crate and hence not published on [crates.io](crates.io). However, since a release tag exists for the crate, `release-plz` tries to download the latest release from [crates.io](crates.io) and fails. See [this failed release pipeline run](https://github.com/dfinity/sol-rpc-canister/actions/runs/17233112571) for example.

I've opened [a PR](https://github.com/release-plz/release-plz/pull/2357) to fix this issue in the upstream release-plz repository, and updated the release pipeline here to build the release-plz GitHub action from source to include those changes. This is a temporary solution until the changes are merged in the upstream repository. The release pipeline runs again successfully when using the patched release-plz GitHub action (see [this run](https://github.com/dfinity/sol-rpc-canister/actions/runs/17321578196)).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,16 +19,12 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@v1
-        with:
-          toolchain: nightly
       - name: Install release-plz from fork
         run: |
           git clone https://github.com/lpahlavi/release-plz.git /tmp/release-plz-fork
           cd /tmp/release-plz-fork
           git checkout 9b511999ed04cc5070fe5efcd5481b6b2cc035c1
-          cargo +nightly install --path crates/release_plz
+          cargo +1.88.0 install --path crates/release_plz
       - name: Run release-plz
         run: release-plz release-pr --git-token ${{ secrets.GITHUB_TOKEN }}
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,6 @@ jobs:
           git checkout bd11ce7ef1683c1b4dd359b57c867cfdf3e04f4a
           cargo +nightly install --path crates/release_plz
       - name: Run release-plz
-        run: release-plz release-pr
+        run: release-plz release-pr --git-token ${{ secrets.GITHUB_TOKEN }}
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


lpahlavi:
@gregorydemay See the updated PR description. I've opened an [issue](https://github.com/release-plz/release-plz/issues/2358) and [PR](https://github.com/release-plz/release-plz/pull/2357) in the upstream release-plz repository and this PR now just uses the patched GitHub action in the meantime until those changes are merged upstream.

.github/workflows/release.yml
```diff
@@ -19,10 +19,17 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Run release-plz
-        uses: release-plz/action@8724d33cd97b8295051102e2e19ca592962238f5 # v0.5.108
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@v1
         with:
-          command: release-pr
+          toolchain: nightly
```
gregorydemay, line 25: do we really need nightly?

gregorydemay (approved):
Thanks for the fix!

.github/workflows/release.yml
lpahlavi: Actually it was a huge pain to get `release-plz` to compile from source (see the commits in this branch 😅) due to their use of some unstable features (I believe `if let` chains in `if` statements). I guess this should hopefully be just a temporary fix so I thought it's good enough, but we can also experiment to find was versions of Rust work. WDYT?

lpahlavi:
> There is one commit that should be picked up if the changelog was automatically generated
> 
> ```
> git log --format='%C(auto) %h %s' sol_rpc_canister-v1.1.0..HEAD -- canister
>  c3d5468 chore: upgrade SOL RPC canister to v1.1.0 (#206)
> ```
> 
> But that's I think not too tragic and could be manually adapted in the generated PR?

@gregorydemay I guess actually `release-plz` is smart enough to ignore this commit since it's just updating the crate version. It's not too much of a hassle, but I mostly wanted to make sure I didn't somehow break something in my upstream PR leading to an incomplete changelog / incorrect version determination.

.github/workflows/release.yml
gregorydemay: If it's just to compile `release-plz`, then I'm totally fine with it, but I'm more worried that nightly could be use to compile the sol-rpc-canister and this would be problematic (for build reproducibility)

PR #217, opened 2025-08-28
test: local ICP Ninja example deployment

lpahlavi:
(XC-454) Test local deployment of ICP Ninja `basic_solana` example. This simulates what a user would do when downloading their project from ICP Ninja and deploying it locally. This includes downloading the latest SOL RPC canister WASM file from GitHub and changing the value of `solana_network ` in the `init_args` in `dfx.json` from `Devnet ` to `Custom = record { url = \"https://api.devnet.solana.com\"; headers = null }` as outlined in the `basic_solana` README.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,10 +140,10 @@ jobs:
           dfx deploy
           out=$(dfx canister call basic_solana solana_account --output json)
           echo "$out"
-          echo "$out" | grep -E "[1-9A-HJ-NP-Za-km-z]{32,44}" > /dev/null || { echo "❌ Call to 'solana_account' failed"; exit 1; }
+          echo "$out" | grep -Eq "[1-9A-HJ-NP-Za-km-z]{32,44}" || { echo "❌ Call to 'solana_account' failed"; exit 1; }
           out=$(dfx canister call basic_solana get_balance --output json)
           echo "$out"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -114,6 +114,37 @@ jobs:
...
+        working-directory: examples/basic_solana/ninja
+        run: |
+          dfx start --clean --background
+          dfx deploy
+          out=$(dfx canister call basic_solana solana_account --output json)
+          echo "$out"
+          echo "$out" | grep -E "[1-9A-HJ-NP-Za-km-z]{32,44}" > /dev/null || { echo "❌ Call to 'solana_account' failed"; exit 1; }
+          out=$(dfx canister call basic_solana get_balance --output json)
+          echo "$out"
+          echo "$out" | grep -E "[0-9]+(?:_[0-9]+)*" > /dev/null || { echo "❌ Call to 'get_balance' failed"; exit 1; }
```
gregorydemay, line 146: nit: there might be a problem with the grep since the [job](https://github.com/dfinity/sol-rpc-canister/actions/runs/17296774166/job/49096838702?pr=217) issues a warning
```
grep: warning: ? at start of expression
```

gregorydemay (approved):
Thanks a lot @lpahlavi for this PR!

.github/workflows/ci.yml
lpahlavi: Very good catch! Turns out this is PCRE regex syntax for a non-capturing group, but we need POSIX syntax which doesn't support that. I've removed it (also in the `icp-ninja-tests` step) as it's unnecessary for us.

PR #223, opened 2025-09-02
ci: re-enable `check-ninja-cargo-toml` in CI pipeline

lpahlavi:
(XC-454) Since #222 was merged, re-enable the CI pipeline step to ensure the `Cargo.toml` for the ICP Ninja deployment of `basic_solana` is up-to-date, and group it with the other `icp-ninja-tests` added in #217.


gregorydemay (approved):
Thanks for re-enabling that @lpahlavi !

PR #237, opened 2025-09-24
test: integration tests that RPC config is respected

lpahlavi:
This PR adds integration tests to ensure that RPC config parameters are correctly respected when invoking the various SOL RPC canister endpoints. It also introduces a missing integration test for `getTransaction`.

### DIFF
--- a/integration_tests/tests/tests.rs
+++ b/integration_tests/tests/tests.rs
@@ -10,7 +10,8 @@ use serde_json::{json, Value};
 use sol_rpc_canister::constants::*;
 use sol_rpc_client::{RequestBuilder, SolRpcClient, SolRpcConfig, SolRpcEndpoint};
 use sol_rpc_int_tests::{
-    mock::MockOutcallBuilder, PocketIcRuntime, Setup, SolRpcTestClient, DEFAULT_CALLER_TEST_ID,
+    json_rpc_sequential_id, mock::MockOutcallBuilder, PocketIcRuntime, Setup, SolRpcTestClient,
+    DEFAULT_CALLER_TEST_ID,
 };
 use sol_rpc_types::{
     CommitmentLevel, ConfirmedTransactionStatusWithSignature, ConsensusStrategy,
@@ -1268,7 +1269,6 @@ mod cycles_cost_tests {
 
 mod rpc_config_tests {
     use super::*;
-    use sol_rpc_int_tests::json_rpc_sequential_id;
 
     #[tokio::test]
     async fn should_respect_response_size_estimate() {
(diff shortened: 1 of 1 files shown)


integration_tests/tests/tests.rs
```diff
@@ -170,24 +175,7 @@ mod get_account_info_tests {
...
-                                "executable": false,
-                                "lamports": 88849814690250u64,
-                                "owner": "11111111111111111111111111111111",
-                                "rentEpoch": 18446744073709551615u64,
-                                "space": 0
-                            }
-                        },
-                    }),
-                )
+                .mock_sequential_json_rpc_responses::<3>(200, get_account_info_response(first_id))
```
gregorydemay, line 178: Small ask: for the future, if it's not possible to separate refactoring from the current PR, please do it in separate commits. Otherwise reviewing this is annoying.

integration_tests/tests/tests.rs
```diff
@@ -401,17 +368,13 @@ mod get_slot_tests {
...
                 .enumerate()
                 .map(|(id, slot)| {
                     MockOutcallBuilder::new(
                         200,
-                        json!({
-                            "id": Id::from(ConstantSizeId::from(id as u64 + first_id as u64)),
-                            "jsonrpc": "2.0",
-                            "result": slot,
-                        }),
+                        get_slot_response_with_slot(id as u8 + first_id, slot),
```
gregorydemay, line 377: nit: `slot_response_with_slot` is a bit puzzling. I think the slot should always be explicit (since afterwards there should be an assertion that the response value is the correct one). So I would just keep only one method `get_slot_response`.

integration_tests/tests/tests.rs
```diff
@@ -1624,53 +1049,353 @@ mod cycles_cost_tests {
...
+                MultiRpcResult<CandidOutput>,
+                MultiRpcResult<Output>,
+            >,
+            Config: CandidType + Clone + Send + SolRpcConfig,
+            Params: CandidType + Clone + Send,
+            CandidOutput: CandidType + DeserializeOwned,
+            Output: Debug + PartialEq,
+            MultiRpcResult<CandidOutput>: Into<MultiRpcResult<Output>>,
+        {
+            // Expect multiple retries with increasingly larger values of `max_response_bytes`
```
gregorydemay, line 1298: It's not clear to me why we need to test the retries from 1 byte to 2MB and then the no retries when the max response size is already at 2MB. Why not just 1 particular large value, e.g. 1_999_999 and then check via the mocks that this is indeed the set response size?

libs/client/src/request/mod.rs
```diff
@@ -783,6 +783,85 @@ impl<Runtime, Config, Params, CandidOutput, Output>
     }
 }
 
+/// Common behavior for the RPC config for SOL RPC canister endpoints.
+pub trait SolRpcConfig: Default {
```
gregorydemay, line 787: I don't have the impression the implementations of this trait need the `Default` trait bound, no?

libs/client/src/request/mod.rs
```diff
@@ -783,6 +783,85 @@ impl<Runtime, Config, Params, CandidOutput, Output>
...
+
+impl<Runtime, Config: SolRpcConfig, Params, CandidOutput, Output>
+    RequestBuilder<Runtime, Config, Params, CandidOutput, Output>
+{
+    /// Change the response size estimate to use for that request.
+    pub fn with_response_size_estimate(mut self, response_size_estimate: u64) -> Self {
+        self.request.rpc_config = Some(
+            self.request
+                .rpc_config
+                .unwrap_or_default()
```
gregorydemay, line 847: Here default could be added as an additional trait bound on `Config`.

gregorydemay (approved):
Thanks a lot @lpahlavi for this PR and sorry for the delay in reviewing it. Couple of minor comments but otherwise LGTM!

integration_tests/tests/tests.rs
lpahlavi: Good point. Done.

PR #238, opened 2025-09-26
ci: check canister endpoints with `ic-wasm`

lpahlavi:
This PR adds a step to the CI pipeline that runs `ic-wasm check-endpoints` to verify that there are no accidentally or maliciously exported canister endpoints in the canister's WASM.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,11 +151,11 @@ jobs:
           name: sol_rpc_canister.wasm.gz
 
       - name: 'Run IC WASM check-endpoints'
-        uses: dfinity/ic-wasm@d2d1f16f8636891d9a7da50862efa7f8535efcbd
+        uses: dfinity/ic-wasm@72c486f152273f549bcdca50b2297be4a7e0576e
         with:
           command: check-endpoints
           wasm: ${{ github.workspace }}/sol_rpc_canister.wasm.gz
-          hidden: ./canister/hidden_endpoints.txt
+          args: --hidden ./canister/hidden_endpoints.txt
 
   integration-tests:
     needs: [ reproducible-build ]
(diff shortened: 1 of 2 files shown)


canister/hidden_endpoints.txt
```diff
@@ -0,0 +1,14 @@
+canister_query:cleanup_response
+canister_query:http_request
+canister_query:verifyApiKey
+canister_init
+canister_post_upgrade
+main
+rust_zstd_wasm_shim_calloc
```
gregorydemay, line 7: nit: could we maybe add a comment that this comes from that library `"solana-account-decoder-client-types/zstd"`

.github/workflows/ci.yml
```diff
@@ -138,6 +138,27 @@ jobs:
...
+          name: sol_rpc_canister.wasm.gz
+
+      - name: 'Install `ic-wasm`'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: ic-wasm
+
+      - name: 'Check canister endpoints'
+        run: |
+          ic-wasm "$GITHUB_WORKSPACE/sol_rpc_canister.wasm.gz" check-endpoints --hidden ./canister/hidden_endpoints.txt
```
gregorydemay, line 160: nice 😎 . Could you also add this to the EVM RPC canister 🙏

gregorydemay (approved):
Thanks a lot @lpahlavi for this PR!

canister/hidden_endpoints.txt
lpahlavi: Unfortunately comments aren't currently supported in the "hidden endpoints" files. I realized when creating this file here that it could have been a nice feature... I'll create another PR in `ic-wasm` and then add the comments here (and in the EVM RPC canister if necessary) once that is merged.

.github/workflows/ci.yml
lpahlavi: Yes! Here is the PR: [dfinity/evm-rpc-canister #448](https://github.com/dfinity/evm-rpc-canister/pull/488)

PR #320, opened 2026-05-08
ci: pin ic-wasm to 0.9.10

gregorydemay:
## Summary

Pin `ic-wasm` to a specific version (`0.9.10` — the latest published on crates.io) in the `check-canister-wasm-endpoints` job. Without a version pin, the install resolves to the latest release at CI run time, so a future `ic-wasm` release can silently change the meaning of the check on an unrelated PR.


mbjorkqvist (approved):
Thanks @gregorydemay! Nit: The "Note on CI" comment in the PR description is now outdated and could be removed.

PR #321, opened 2026-05-08
build(deps): revert all post-93a385b dependency bumps

gregorydemay:
## Summary

Restore `Cargo.toml` and `Cargo.lock` to their state at commit `93a385b` (2026-04-29, the last commit on `main` where CI passed). Six Dependabot bumps merged in quick succession on the morning of 2026-04-29 left `main` red because **branch protection on this repo doesn't currently gate on status checks** (only on approval + CODEOWNERS), so each bump merged without ever passing CI:

- #311 `pocket-ic` 12.0.0 → 13.0.0
- #312 `solana-message` 3.1.0 → 4.1.1 — removes `Message::serialize()`
- #314 `bincode` 1.3.3 → 3.0.0 — `compile_error!` sentinel (https://xkcd.com/2347/)
- #315 `ic-management-canister-types` 0.5.0 → 0.7.1
- #316 `ic-agent` 0.46.2 → 0.47.2
- #317 `solana-reward-info` 3.0.0 → 6.2.0 — adds `DeactivatedStake` variant

Reverting all six in one go restores `main` to the known-green state. Each upgrade can be re-applied individually once branch protection is configured to require CI to pass.

## Branch protection follow-up (out of scope here)

`gh api repos/dfinity/sol-rpc-canister/branches/main/protection` returns 404 (not protected); the repo uses org-level rulesets instead, but none of them lists a `required_status_checks` rule. As a result, anyone with merge rights can land a PR with red CI. Adding a `required_status_checks` rule (org or repo level) listing `lint`, `unit-tests`, `cargo-doc`, `check-cargo-lock`, `reproducible-build`, and `check-canister-wasm-endpoints` would prevent recurrence. This requires admin rights so it's deliberately not part of this PR.

PR #322, opened 2026-05-08
ci: add checks-pass aggregator job

gregorydemay:
## Summary

Add a single `checks-pass` aggregator status check that depends on every quality-gate job and explicitly fails if any upstream job is non-success (failure, cancelled, or skipped).

This lets GitHub branch protection / rulesets gate merges on a single context name (`checks-pass`) instead of enumerating every individual job. It also future-proofs the protection rule: adding or removing a CI job only needs an update to the `needs:` list here.

Mirrors the pattern already in use in `dfinity/dogecoin-canister` (`.github/workflows/workflow.yml` `checks-pass`).

## Ordering

Should merge **after #321** (the bulk dep revert that fixes `main`). CI on this PR is expected to be red until #321 lands, since `main` itself currently fails on `unit-tests`/`reproducible-build`/`cargo-doc`/etc. The new aggregator simply codifies that "if any of those is red, the merge gate is closed."

## Follow-up (admin action, not in this PR)

After this merges to `main`, an admin should add a `required_status_checks` rule (in `Settings > Rules` or via the org-level rulesets) listing `checks-pass` as a required context. That closes the gap that allowed PRs #311–#318 to merge with red CI and break `main` for ten days.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,3 +288,52 @@ jobs:
         env:
           DFX_DEPLOY_KEY: ${{ secrets.DFX_DEPLOY_KEY }}
         run: cargo test --locked --package sol_rpc_e2e_tests -- --test-threads 1 --nocapture
+
+  # Single aggregator status check. Configure GitHub branch protection
+  # to require `checks-pass` so that any failing or skipped upstream
+  # job blocks the merge, without having to enumerate every individual
+  # job in the ruleset.
+  checks-pass:
+    # Always run this job!
+    if: always()
+    needs:
+      [
+        lint,
+        check-cargo-lock,
+        cargo-doc,
+        reproducible-build,
+        unit-tests,
+        basic-solana-deployment,
+        check-canister-wasm-endpoints,
+        integration-tests,
+        end-to-end-tests,
+      ]
+    runs-on: ubuntu-latest
+    steps:
+      - name: check lint result
+        if: ${{ needs.lint.result != 'success' }}
+        run: exit 1
+      - name: check check-cargo-lock result
+        if: ${{ needs.check-cargo-lock.result != 'success' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -288,3 +288,29 @@ jobs:
...
+        unit-tests,
+        basic-solana-deployment,
+        check-canister-wasm-endpoints,
+        integration-tests,
+        end-to-end-tests,
+      ]
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          if [[ "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}" == "true" ]]; then
```
mbjorkqvist: nit: The "any failing *or skipped* upstream job blocks the merge" comment suggests that this should also contain `contains(needs.*.result, 'skipped')`. Currently, there's no scenario where a job may be skipped with no other jobs failing, so this is not strictly needed, but it may make sense to clarify the comment, or for robustness add a check for `skipped` also, if it is possible that such a scenario may be added in the future, and we wouldn't want `checks-pass` to silently ignore it.

mbjorkqvist (approved):
Thanks @gregorydemay, just one minor nit!

.github/workflows/ci.yml
gregorydemay: Good catch — addressed in 85287fb by adding `|| contains(needs.*.result, 'skipped')` to the condition. As you note, today no job has an `if:` that could skip on its own, but the explicit check is defensive against a future job that opts out that way.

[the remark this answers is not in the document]
gregorydemay: Addressed in 85287fb — `|| contains(needs.*.result, 'skipped')` is now in the condition, matching the docstring.

PR #375, opened 2026-07-23
chore(examples): migrate basic_solana from dfx to icp-cli

marc0olo:
Migrates the `basic_solana` example from a multi-`dfx` setup to **icp-cli**, following the pattern from [`basic_ethereum`](https://github.com/dfinity/examples/tree/master/rust/basic_ethereum). The three per-environment dfx setups collapse into a single `icp.yaml`; there is no separate `ninja/` folder.

## What changed

- **Single `icp.yaml`** with `local` and `ic` environments replaces the `local/`, `mainnet/`, and `ninja/` folders and `build.sh`. `local` deploys `backend` + a pre-built `sol_rpc` (`sol_rpc_canister-v1.3.0`); `ic` deploys only `backend` and points it at the shared mainnet SOL RPC canister.
- **Crate restructured** into a `backend/` `cdylib` workspace member; `src/main.rs` + `src/lib.rs` are merged into `backend/lib.rs`. The hand-written `basic_solana.did` is removed — the Candid interface is generated via `ic_cdk::export_candid!`. A generated `backend/backend.did` is committed and guarded by a compatibility test (see below).
- **SOL RPC canister ID via the `PUBLIC_CANISTER_ID:sol_rpc` canister environment variable** (auto-injected locally after deploying the pre-built `sol_rpc` canister; set to the shared mainnet canister `tghme-zyaaa-aaaar-qarca-cai` on `ic`). Removed from `InitArg`.
- **Threshold Ed25519 key name is now a plain string** (`InitArg.ed25519_key_name : opt text`), accepting only `"test_key_1"` (default) or `"key_1"`; `dfx_test_key` and the `Ed25519KeyName` enum are gone.
- Added `.cargo/config.toml` (getrandom wasm backend cfg, needed now that the Solana SDK pulls getrandom 0.3) and `rust-toolchain.toml`; crates bumped to the repository's current versions (`sol_rpc_client` 6.0.0, etc.). `get_recent_block()` replaces the removed `estimate_recent_blockhash()`.
- **Docs**: README rewritten for the icp-cli workflow, including a one-click **ICP Ninja** deploy section and the macOS LLVM requirement (needed to compile `zstd` for wasm).

## Tests / CI

- The **PocketIC integration test** is kept (same coverage). `backend` is a `cdylib` (no `bin`), so it can't be built via `load_wasm`'s `--bin`; the test builds the cdylib directly (or uses `BACKEND_WASM_PATH` in CI) and injects the SOL RPC ID via `CanisterSettings.environment_variables`, which requires **PocketIC 15** — so it runs in its own CI job, while `sol_rpc_int_tests` stays on PocketIC 13.
- **Candid interface compatibility test** restored: `candid_interface_matches_committed_did` fails if the exported interface drifts from `backend/backend.did` (regenerate with `candid-extractor`).
- **CI** reworked for icp-cli: deploy + smoke test on Linux and macOS, a dedicated PocketIC-15 integration job, and no more `dfx`. candid-extractor is fetched with `cargo-binstall`; the getrandom cfg is passed via `RUSTFLAGS` (the workflow's global `-Dwarnings` otherwise overrides `.cargo/config.toml`); `ICP_CLI_GITHUB_TOKEN` authenticates icp-cli's GitHub API calls.

## Points worth a close look

- **`icp.yaml`** — the `local`/`ic` split, `init_args`, and the env-var injection.
- **`backend/lib.rs`** — `sol_rpc_id()` (env var), the `ed25519_key_id(&str)` guard, and `get_recent_block()`.
- **`backend/tests/tests.rs`** — cdylib Wasm loading and the env-var-based SOL RPC wiring.

## Verification

Verified locally: `icp deploy` (local) with live `solana_account` / `get_balance` / `send_sol` against Solana Devnet, the full PocketIC integration test (send / SPL / durable nonce), and `cargo check --locked`. CI covers the Linux + macOS deploys, the PocketIC-15 integration job, and the mainnet end-to-end flow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mbjorkqvist (approved):
Thanks @marc0olo!

